Issue #54, opened 2016-04-29
HdfsRepository supports


liningalex:
I've implemented it so clients can retrieve schema in HDFS.

PR #55, opened 2016-04-29
HdfsRepository supports #54

liningalex:
Hello, 

With this change, Hadoop and Spark apps can retrieve schema in hdfs.

Regards

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -44,8 +44,10 @@ object Dependencies {
     val scalaz7          = "org.scalaz"                 %% "scalaz-core"             % V.scalaz7
     val collUtils        = "com.twitter"                %% "util-collection"         % V.collUtils
     // hadoop hdfs
-    val hadoopCommon     = "org.apache.hadoop"         %  "hadoop-common"            % V.hadoop
-    val hadoopHdfs     = "org.apache.hadoop"         %  "hadoop-hdfs"                % V.hadoop
+    val hadoopCommon     = "org.apache.hadoop"         %  "hadoop-common"            % V.hadoop		 % "provided"
+    val hadoopHdfs       = "org.apache.hadoop"         %  "hadoop-hdfs"              % V.hadoop		 % "provided"
+    val hadoopCommonTest = "org.apache.hadoop"         %  "hadoop-common"            % V.hadoop		 % "test"
+    val hadoopHdfsTest   = "org.apache.hadoop"         %  "hadoop-hdfs"              % V.hadoop		 % "test"
     // Scala (test only)
     val specs2           = "org.specs2"                 %% "specs2"                  % V.specs2          % "test"
     val scalazSpecs2     = "org.typelevel"              %% "scalaz-specs2"           % V.scalazSpecs2    % "test"
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -30,6 +30,7 @@ object Dependencies {
     val specs2          = "2.3.13" // Downgrade to prevent issues in job tests. WAS: "2.3.11"
     val scalazSpecs2    = "0.2"
     val mockito         = "1.10.19"
+    val hadoop          = "2.2.0"
   }
 
   object Libraries {
@@ -42,6 +43,9 @@ object Dependencies {
     val json4sScalaz     = "org.json4s"                 %% "json4s-scalaz"           % V.json4s
     val scalaz7          = "org.scalaz"                 %% "scalaz-core"             % V.scalaz7
     val collUtils        = "com.twitter"                %% "util-collection"         % V.collUtils
+    // hadoop hdfs
+    val hadoopCommon     = "org.apache.hadoop"         %  "hadoop-common"            % V.hadoop
+    val hadoopHdfs     = "org.apache.hadoop"         %  "hadoop-hdfs"                % V.hadoop
     // Scala (test only)
     val specs2           = "org.specs2"                 %% "specs2"                  % V.specs2          % "test"
     val scalazSpecs2     = "org.typelevel"              %% "scalaz-specs2"           % V.scalazSpecs2    % "test"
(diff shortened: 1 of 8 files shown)


alexanderdean:
Hi @liningalex - thanks for this! This is a super-interesting contribution. Please can you sign our CLA: https://github.com/snowplow/snowplow/wiki/CLA

liningalex:
Signed. Thanks.

alexanderdean:
Confirmed! Thanks so much. Assigning to @chuwy for review.

chuwy:
Hello @liningalex! This is a very interesting. Could you please describe your use-case, I mean why do you need to store Schemas in HDFS?

liningalex:
Hi Anton, it's for big data applications - hadoop map-reduce and spark
where hdfs is the file systems. Thanks. Lining

chuwy:
Thank you @liningalex. Just to clarify my confusion: running Spark/Hadoop jobs are currently able to make requests to schema registries, and what is more important working Hadoop nodes cache these Schemas, so we don't have massive streams of HTTP requests as one may think. And moreover, we're using Iglu Client from Hadoop as well, so this is primary of purpose of Iglu Client - work on Hadoop cluster.

On the other hand: we're trying to make Iglu Client small and slick, so it could be embedded to big variety of applications and including huge dependencies such as hadoop can break its agility. So we need solid reasons to include dependencies like Hadoop. Thanks.

alexanderdean:
Thanks @liningalex for explaining that. I'm wondering if we should do something where we explicitly keep the Hadoop dependencies out of the pom - i.e. we assume that if somebody is using HDFS to store their schemas, their code is already bringing in the Hadoop dependencies. Thoughts @chuwy ?

chuwy:
@alexanderdean I believe in that very rare (from my point of view of course) case where someone need to store Schemas in HDFS he can also easily implement `RepositoryRef` trait in his app. This deliver us from dealing with dependency complications as well as including `hdfs` in [resolver-config schema](https://github.com/snowplow/iglu-scala-client/blob/master/src/main/resources/iglu-client-embedded/schemas/com.snowplowanalytics.iglu/resolver-config/jsonschema/1-0-1) (which is btw missing in this PR) because it can be created programmatically rather than parsing resolver config. I hope that makes sense.

alexanderdean:
Hey @chuwy - it feels a shame not to include this functionality just because of the dependency requirement. Why wouldn't my suggested approach of just excluding the Hadoop dependency work?

chuwy:
Hello @liningalex. We decided that having this feature in Iglu Client can be really handy. To actually merge this PR we need to do a couple more things:
* Add `hdfs` to [resolver schema](https://github.com/snowplow/iglu-scala-client/tree/master/src/main/resources/iglu-client-embedded/schemas/com.snowplowanalytics.iglu/resolver-config/jsonschema). This is used when we invoke `Resolver.parse(jsonConfig)` to create a resolver which is default way to use it. Otherwise we can add HDFS repository to Iglu only directly using constructors.
* Make a PR to [Iglu Central](https://github.com/snowplow/iglu-central) with same Schema, so it can be available publicly.

Thank you very much.

liningalex:
Hi Anton, Done. please review. Thanks. Lining

On Tue, May 10, 2016 at 2:18 AM, Anton Parkhomenko <notifications@github.com

chuwy:
Hey @liningalex!

We need just few minor changes to make this PR complete:

- [x] Make sure you added 1-1-0 Schema from iglu-central to embedded repo (in main/resources/...)

- [ ] Write unit test to parse `HdfsRepository` (just like this [one](https://github.com/snowplow/iglu-scala-client/blob/master/src/test/scala/com.snowplowanalytics.iglu.client/ResolverSpec.scala#L89)), so we can be sure we didn't miss anything in new `1-1-0` Schema

- [ ] Most important: exclude hadoop dependencies from POM. You can do this using "provided" scope as described [here](http://stackoverflow.com/a/27836766/998523). With this scope Iglu Client will not bring huge Hadoop dependencies to applications where Hadoop isn't used. 

- [ ] Please, use fully-qualified imports like `import org.apache.hadoop.fs.{ Path, FileSystem }` and omit semicolons, they're not necessary in Scala. Also, it seems `hadoop.io._` and `hadoop.utils._` are not used anywhere?

Thanks!  And sorry for being so pedantic.

alexanderdean:
Thanks @chuwy

liningalex:
Hi Anton,
We don't have a hdfs://iglucentral.com/, wondering how do we do
HdfsRepository ResolverSpec test?
If we exclude Hadoop dependencies in POM, how do we resolve unit testing
failure?

Thanks

chuwy:
Hello @liningalex 
1. We don't need `hdfs://iglucentral.com/` for this test, we need just to be sure `Resolver.parse` method can create valid `Resolver` instance from full [resolver_config.json](https://github.com/snowplow/iglu/wiki/Iglu-client-configuration#standard-configuration) (with both old HTTP repository and new HDFS). Like you're doing [here](https://github.com/snowplow/iglu-scala-client/pull/55/files#diff-eee9ecc7325fe4e3931cc5e00db72b00R58), but your unit-test checks only `HdfsRepositoryRef.parse` which doesn't use new embedded Schema. `Resolver.parse` on the other hand uses embedded repo and if we've done something wrong in new Schema - test will help us to find out.
2. Hm, that is interesting. I guess, we can try to add test-scoped Hadoop dependency like: 
```scala
val hadoopCommon         = "org.apache.hadoop"         %  "hadoop-common"            % V.hadoop   % "provided"
val hadoopCommonTest     = "org.apache.hadoop"         %  "hadoop-common"            % V.hadoop   % "test"
...
```
I think that should work.

liningalex:
Hi Anton, I pushed the changes, would you review? thanks.

chuwy:
Hello @liningalex,

This looks good to me! I'm scheduling it to next release.
/cc @alexanderdean 

Thank you again!

chuwy:
Pushing back.

alexanderdean:
Happy to keep open...

chuwy:
Closing as it's too far from current work tree.

PR #89, opened 2018-07-13
Scalafmt and bumps

szareiangm:
I will add more library bumps to it as I explore the upgrades possible. 
This PR is large because I ran below commands for scalafmt:
```
compile
test:scalafmt
sbt:scalafmt
```
Please review it per commits. 
The ultimate goal is to prepare the codebase for Scala 2.12. 

CC: @nafid @zakipatel

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -21,7 +21,7 @@ object Dependencies {
     // Scala
     val json4s          = "3.2.11"
     val scalaz7         = "7.0.9"
-    val collUtils       = "6.20.0"
+    val collUtils       = "18.7.0"
     // Scala (test only)
     val specs2          = "2.3.13" // Downgrade to prevent issues in job tests. WAS: "2.3.11"
     val scalazSpecs2    = "0.2"
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -23,7 +23,7 @@ object Dependencies {
     val scalaz7         = "7.0.9"
     val collUtils       = "18.7.0"
     // Scala (test only)
-    val specs2          = "2.3.13" // Downgrade to prevent issues in job tests. WAS: "2.3.11"
+    val specs2          = "4.3.2"
     val scalazSpecs2    = "0.2"
     val mockito         = "1.10.19"
   }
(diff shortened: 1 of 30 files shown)


chuwy:
Thanks for all your work, @szareiangm. Please let me know when you're done and I'll review it. Also please pay attention to comments in #90 and #91.

szareiangm:
I think that these were the last things we could do before adding support for 2.12. Beucause what I see now is the error below when I switch the compile to 2.12:
```
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: org.json4s#json4s-scalaz_2.12;3.2.11: not found
[warn] 	:: org.specs2#specs2_2.12;2.3.13: not found
[warn] 	:: org.typelevel#scalaz-specs2_2.12;0.2: not found
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
```

I should check the tickets to convert `Scalaz->Cats` and `Json4s->Circe`.  They should be done before we could add Scala 2.12.

@chuwy  You can go ahead with the review for this PR.

chuwy:
Hey @szareiangm! I'll do review tomorrow morning. Meanwhile, regarding those errors: all libraries except `specs2` itself do not provide much functionality. All of them can refactored into something as simple as `toEither`. `specs2` itself can be bumped to 4.0.3. 4.x branch does not use scalaz as dependency.

project/Dependencies.scala
```diff
@@ -21,9 +21,9 @@ object Dependencies {
     // Scala
     val json4s          = "3.2.11"
     val scalaz7         = "7.0.9"
-    val collUtils       = "6.20.0"
```
chuwy: I'm bit hazy about such a huge leap. Chances are big that other libraries are not 18-compatible yet which means we can hit runtime exceptions. If there's a need for this bump I'd like to first check if scala-common-enrich is compatible with it.

szareiangm: I won't be insisting on collection-utils. The only reason was that there was a big gap and I just wanted to catch up.
I will doing some work on `scala-common-enrich` this week as part of optional enrichment work. I can check the compatibility if you want.

snowplowcla:
@szareiangm  has signed the Software Grant and Corporate Contributor License Agreement

BenFradet:
@chuwy doesn't have to the case no

project/Dependencies.scala
chuwy: Yes, MiMa is quite wide-spread. I was thinking about it the other day, but decided that we have not that big code estate to have troubles with maintenance. On the other hand we have the clear evidence that outside contributors are not always aware of dependency tree and we have few internal issues with it, so it might be a good idea to reconsider. What do you think @BenFradet?

BenFradet: yes we will have to at some point, if we can rack up some xp on a guinea pig project (i.e. this one) that'd be great.

szareiangm: That's exciting to hear. Helps a lot for moving forward with the caution needed. 🙌  ✌️

asoltysik:
@szareiangm Thanks a lot, your work on this PR is included in #99

PR #105, opened 2018-09-04
Migrate to Circe, add Iglu Core, use scala-lrumap

asoltysik:
PR number 3 (still pretty big, should I split this further?)


src/main/scala/com.snowplowanalytics.iglu/client/Resolver.scala
```diff
@@ -295,26 +288,26 @@ case class Resolver(
...
-  /**
-   * Our LRU cache.
-   */
+  // TODO: ideally replace unsafeRunSync() with referentially transparent API
   object cache {
 
     private val lru: Option[SchemaLruMap] =
-      if (cacheSize > 0) Some(new SchemaLruMap(cacheSize)) else None
+      if (cacheSize > 0)
+        Some(LruMap.create[IO, SchemaKey, SchemaLookupStamped](cacheSize).unsafeRunSync())
```
asoltysik: We either keep the non-RT API in 0.6 with `unsafeRunSync()` calls in cache or we need to refactor the whole API to be referentially transparent.

src/main/scala/com.snowplowanalytics.iglu/client/repositories/EmbeddedRepositoryRef.scala
```diff
@@ -51,31 +51,19 @@ object EmbeddedRepositoryRef {
    * @return true if this is the configuration for
    *         an EmbeddedRepositoryRef, else false
    */
-  def isEmbedded(config: JValue): Boolean =
-    (config \ "connection" \ "embedded").toSome.isDefined
+  def isEmbedded(config: Json): Boolean =
+    root.connection.embedded.json.getOption(config).isDefined
```
asoltysik: Are you guys okay with *optics* dependency?

src/main/scala/com.snowplowanalytics.iglu/client/Resolver.scala
chuwy: I vote for fully-RT API. Mixing eager and strict evaluations is just an another huge source of bugs as we saw in Scala Tracker.

src/main/scala/com.snowplowanalytics.iglu/client/repositories/EmbeddedRepositoryRef.scala
chuwy: Not that I'm really against it (though I am, sorry), but full `ResolverConfig` case class with associated decoder would be much-much better, we don't want pass around a Json, but full correct config.

src/main/scala/com.snowplowanalytics.iglu/client/utils/SchemaKeyUtils.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import cats.implicits._
+import com.snowplowanalytics.iglu.core.SchemaKey
+
+import validation.ProcessingMessage
+
+private[client] object SchemaKeyUtils {
+
+  def toPath(prefix: String, key: SchemaKey): String =
+    s"$prefix/schemas/${key.vendor}/${key.name}/${key.format}/${key.version.asString}"
```
chuwy: It is `s"$prefix/schemas/${key.asPath}"`

asoltysik: Hmm can't find such method https://github.com/snowplow/iglu/blob/master/0-common/scala-core/src/main/scala/com.snowplowanalytics.iglu/core/SchemaKey.scala

src/main/scala/com.snowplowanalytics.iglu/client/Resolver.scala
```diff
@@ -12,50 +12,50 @@
...
-import cats.instances.map._
-import cats.instances.option._
-import cats.syntax.apply._
-import cats.syntax.applicativeError._
-import cats.syntax.semigroup._
-import cats.syntax.option._
-import cats.syntax.traverse._
-import cats.syntax.validated._
-import cats.data.NonEmptyList
+import cats.implicits._
```
asoltysik: It was too much of a burden to maintain this import list

asoltysik:
@BenFradet publishing a new version of `scala-lru-map` with https://github.com/snowplow-incubator/scala-lru-map/pull/10 would allow us to depend on `cats-effect` 1.0.0.

src/main/scala/com.snowplowanalytics.iglu/client/SchemaCache.scala
```diff
@@ -0,0 +1,93 @@
...
+    ttl.fold(true)(ttlVal => unixSeconds - storedTstamp < ttlVal)
+
+}
+
+object SchemaCache {
+
+  def apply[F[_]: Sync](size: Int, ttl: Option[Int] = None): F[SchemaCache[F]] = {
+    Sync[F].unit
+      .ensure(new IllegalArgumentException("Cache size must be greater than 0"))(_ => size > 0)
+      .ensure(new IllegalArgumentException("Cache TTL must be greater than 0"))(_ =>
```
chuwy: `ttl.exists(_ > 0)` is much easier to read. Also, I actually don't understand why we have "greater than 0" constraints. I believe we should have "greater or equal" as 0 looks like valid value for me ("no cache").

src/main/scala/com.snowplowanalytics.iglu/client/Resolver.scala
```diff
@@ -90,75 +90,68 @@ object Resolver {
...
-          )
-        (cacheSize, repositoryRefs, cacheTtl).mapN {
-          Resolver(_, _, _)
-        }
-      }
-      .leftMap(errors =>
-        errors.prepend("Resolver configuration failed JSON Schema validation".toProcessingMessage))
+    val resolver = Bootstrap.resolver[F]
+
+    resolver.flatMap { implicit resolver =>
```
chuwy: I would like to get rid of "implicit resolver" pattern. I think we abused Scala implicits a lot with it. I believe it is generally a bad practice to pass resources (such as `ExecutionContext`) implicitly.

asoltysik: yup it is especially cumbersome as the creation of a resolver is now an effect

src/main/scala/com.snowplowanalytics.iglu/client/SchemaCache.scala
asoltysik: It's not the same thing - `None.exists(_ > 0) === false`, `! None.exists(_ <= 0) === true`

asoltysik: @chuwy I've implemented the "no cache" case

https://github.com/snowplow/iglu-scala-client/pull/105/files#diff-02c62df5ff9f8274c5dd567877723cedR87

src/main/scala/com.snowplowanalytics.iglu/client/SchemaCache.scala
```diff
@@ -0,0 +1,94 @@
...
+    for {
+      seconds <- currentSeconds
+      _       <- lru.put(schemaKey, (seconds, schema))
+    } yield schema
+
+  /**
+   * Get Unix epoch timestamp in seconds
+   */
+  private val currentSeconds: F[Int] =
+    Sync[F].delay((System.currentTimeMillis() / 1000).toInt)
```
BenFradet: couldn't we use `Clock` now?

src/main/scala/com.snowplowanalytics.iglu/client/SchemaCache.scala
```diff
@@ -0,0 +1,94 @@
...
+   * @return true if
+   */
+  private def isViable(unixSeconds: Int, storedTstamp: Int): Boolean =
+    ttl.fold(true)(ttlVal => unixSeconds - storedTstamp < ttlVal)
+
+}
+
+object SchemaCache {
+
+  def apply[F[_]: Sync](size: Int, ttl: Option[Int] = None): F[Option[SchemaCache[F]]] = {
```
BenFradet: I wouldn't expect an Option from a smart constructor, what's the reasoning behind this?

src/main/scala/com.snowplowanalytics.iglu/client/repositories/EmbeddedRepositoryRef.scala
```diff
@@ -125,35 +109,34 @@ case class EmbeddedRepositoryRef(override val config: RepositoryRefConfig, path:
...
-          .invalid
-          .toProcessingMessage
-      case e: Throwable =>
-        s"Unknown problem reading and parsing ${schemaPath} in ${descriptor} Iglu repository ${config.name}: ${VE
-          .getThrowableMessage(e)}".invalid.toProcessingMessage
-    } finally {
-      streamOpt.foreach(_.close())
+  def lookupSchema[F[_]: Sync](schemaKey: SchemaKey): F[Either[ProcessingMessage, Option[Json]]] = {
+    val schemaPath = SchemaKeyUtils.toPath(path, schemaKey)
+    val streamOpt = Option(getClass.getResource(schemaPath))
```
BenFradet: isn't this side-effectful?

BenFradet: why the option wrapping also?

src/main/scala/com.snowplowanalytics.iglu/client/repositories/EmbeddedRepositoryRef.scala
```diff
@@ -125,35 +109,34 @@ case class EmbeddedRepositoryRef(override val config: RepositoryRefConfig, path:
...
+            ProcessingMessage(
+              s"Problem parsing $schemaPath as JSON in $descriptor Iglu repository ${config.name}: %s"
+                .format(VE.stripInstanceEtc(failure.message))))
+      } catch {
+        case e: Throwable =>
+          ProcessingMessage(
+            s"Unknown problem reading and parsing ${schemaPath} in ${descriptor} Iglu repository ${config.name}: ${VE
+              .getThrowableMessage(e)}").asLeft
+      } finally {
+        streamOpt.foreach(_.close())
```
BenFradet: couldn't we use `Bracket` for this?

src/main/scala/com.snowplowanalytics.iglu/client/utils/JacksonCatsUtils.scala
```diff
@@ -0,0 +1,27 @@
...
+// Jackson
+import com.fasterxml.jackson.databind.{JsonNode, ObjectMapper}
+
+// circe
+import io.circe.Json
+
+object JacksonCatsUtils {
+
+  // TODO: make it a real conversion
+  def circeToJackson(circeJson: Json): JsonNode =
```
BenFradet: can't we use circe-jackson now?

src/main/scala/com.snowplowanalytics.iglu/client/validation/SchemaValidation.scala
```diff
@@ -37,8 +40,14 @@ object SchemaValidation {
    * @param schema JSON Schema
    * @return list of Processing Messages with log level above warning
    */
-  def getErrors(schema: JsonNode): List[ProcessingMessage] =
-    v4Schema.validate(schema).asScala.toList.map(m => ProcessingMessage(m.getMessage))
+  def getErrors(schema: Json): List[ProcessingMessage] = {
+    val jacksonJson = new ObjectMapper().readTree(schema.noSpaces)
```
BenFradet: circe-jackson should help here too, no?

src/main/scala/com.snowplowanalytics.iglu/client/utils/JacksonCatsUtils.scala
asoltysik: Yup, conversions were merged into circe-jackson 0.10

chuwy:
Closing in favor of https://github.com/snowplow/iglu-scala-client/pull/111

PR #129, opened 2020-01-24
Bump to 0.6.3 with SortedMap

dilyand:


chuwy:
Hi @dilyand! Can you remind me what we decided with it? Shall I close this PR?

dilyand:
I'm not sure we made a decision. We fixed the immediate issue with a custom mapping function in the bad row aggregator. However, I think the long-term plan was still to eventually fix it here as well, to ensure we don't need the custom mapping every time we want to work with bad rows in Beam. If we have another solution, then the PR can be closed.

chuwy:
I _think_ long-term solution was to not use Iglu Client in Bad Rows at all, but as we speak I now think that `SortedMap` in Iglu Client would not hurt anybody and probably will help us to avoid further problems with Beam.

So scheduling for 1.1.0.

chuwy:
Cherry-picked into https://github.com/snowplow/iglu-scala-client/pull/148

PR #153, opened 2021-04-16
Ability to run blocking operations on separate thread pool

istreeter:
This PR is offered as a proof of concept, for discussion and to demo what the client will look like if we continue on this root.  I do not suggest we merge it yet.

This follows [the discussion in Enrich](https://github.com/snowplow/enrich/pull/433) about the need to run blocking operations on a dedicated thread pool, especially for apps with a small primary thread pool, e.g. cats-effect IO apps.  This PR solves the problem by passing around a _wrapper_ around a `cats.effect.Blocker`.  It needs to be wrapped, because `Id` apps do not have a `ContextShift` so they cannot directly call `Blocker.blockOn`.

We have discussed breaking out the `BlockerF` into its own micro-library.  Here I include it as part of of the iglu client library for convenience.

### DIFF
--- a/src/main/scala/com.snowplowanalytics.iglu/client/Client.scala
+++ b/src/main/scala/com.snowplowanalytics.iglu/client/Client.scala
@@ -20,7 +20,7 @@ import io.circe.Json
 
 import com.snowplowanalytics.iglu.core.SelfDescribingData
 
-import resolver.{InitListCache, InitSchemaCache}
+import resolver.{BlockerF, InitListCache, InitSchemaCache}
 import resolver.registries.{Registry, RegistryLookup}
 
 /**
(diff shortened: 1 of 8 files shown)


istreeter:
Some other relevant points to consider....
- In Cats Effect 3, you can run blocking operations on a `Sync`; you don't need a separate `Blocker`
- We want to update to Cats Effect 3 soon anyway, so maybe we should make all breaking changes in one go.

benjben:
Indeed I would be in favor of waiting for cats-effect 3 bump rather than introducing a breaking change in Iglu Client, which is used in many places. Especially because there is a cache and the HTTP calls are once in a while only.

istreeter:
Closing in favour of #161, which will use a http4s client to run non-blocking lookups.

Issue #152, opened 2020-09-21
Explore "exhausted input" errors


chuwy:
I've seen this several times in test environments, but never paid any attention. Recently I saw it for the first time in production pipeline and now we have OSS user mentioning it: https://discourse.snowplowanalytics.com/t/bad-enrichment-invalid-json-exhausted-input/4253

This message comes from circe when it cannot parse a string into valid `Json`, usually with because server responded with empty or truncated string.

However, I'm not sure that this is actually circe's problem - very likey the culprit is the HTTP cilent we use (scalaj) or actual Iglu Server, although I've never seen this with json4s (pre-0.6.0 Iglu clilent)

PR #167, opened 2021-08-05
Release/1.1.1

chuwy:


modules/http4s/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Http4sRegistryLookup.scala
```diff
@@ -94,5 +89,37 @@ object Http4sRegistryLookup {
...
+          RegistryError.ClientFailure(error).asLeft
+        }
+      case response =>
+        response.bodyText.compile.string.map { body =>
+          val error = s"Unexpected response: $body"
+          RegistryError.ClientFailure(error).asLeft
+        }
+    }
+
+    responseResult.attempt.map {
```
istreeter, line 126: I quite like `.handleError` instead of `.attempt` for this kind of thing

modules/http4s/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Http4sRegistryLookup.scala
```diff
@@ -94,5 +97,40 @@ object Http4sRegistryLookup {
       .fromString(s"${cxn.uri.toString.stripSuffix("/")}/schemas/$vendor/$name/jsonschema/$model")
       .leftMap(e => RegistryError.ClientFailure(e.message))
 
+  def runRequest[F[_]: Sync, A: EntityDecoder[F, *]](
+    client: HttpClient[F],
+    req: Request[F]
+  ): F[Either[RegistryError, A]] = {
+    val responseResult = client.run(req).use[F, Either[RegistryError, A]] {
+      case Status.Successful(response) =>
+        response.as[A].map(_.asRight)
```
istreeter, line 106: I think `.as[A]` can raise an error in `F` if the body cannot be parsed as expected.  (I'm only 80% certain on that - need to check).  Is it OK to let the error bubble up, or should this be caught immediately and converted to one of the failure classes?

chuwy: That's a good point. I think likely we are stumbling upon this infamous #152 because the body is trimmed.

modules/http4s/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Http4sRegistryLookup.scala
```diff
@@ -94,5 +99,44 @@ object Http4sRegistryLookup {
...
+  def runRequest[F[_]: Sync, A: EntityDecoder[F, *]](
+    client: HttpClient[F],
+    req: Request[F]
+  ): F[Either[RegistryError, A]] = {
+    val responseResult = client.run(req).use[F, Either[RegistryError, A]] {
+      case Status.Successful(response) =>
+        response
+          .as[A]
+          .map(_.asRight[RegistryError])
+          .handleError {
```
istreeter, line 111: I know this is exactly what I just said to do..... but I'm looking at it again and wondering if it was the right thing 🤦‍♂️.  Because you already handle the error later on at line 134.  And also it's inconsistent, because there are other places where you don't catch errors that might get raised (e.g. `response.bodyText.compile.string`).  I guess you've added an informative error message, so maybe it is better.

Oh also..... I _think_ that methods like `handleError` are guaranteed to be called with non-fatal exceptions only.  So it shouldn't be necessary to pattern match on `NonFatal`.  However, I never found good documentation on this - I think I just read it somewhere in a gitter channel.


I don't suggest changing anything though.  This is all good.

istreeter (approved):

modules/http4s/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Http4sRegistryLookup.scala
chuwy: Yeah, I think there's still a slight difference that would help us to catch trimmed payloads.

On line 134 I catch very IO-ish _client_ exceptions, i.e. network is unavailable, connection pool issue etc. Whereas this `handleError` is specifically about things like decoding failure due trimming or dropped connection and we'll get different error messages.

Although that's post-rationalisation - I also haven't realised the exceptions is caught.

PR #173, opened 2022-02-15
Release/2.0.0

pondzix:


snowplowcla:

project/plugins.sbt
```diff
@@ -1,7 +1,7 @@
 addSbtPlugin("io.github.davidgregory084" % "sbt-tpolecat"    % "0.1.20")
 addSbtPlugin("org.scalameta"             % "sbt-scalafmt"    % "2.4.6")
 addSbtPlugin("com.typesafe"              % "sbt-mima-plugin" % "1.0.1")
-addSbtPlugin("org.scoverage"             % "sbt-scoverage"   % "1.6.1")
+addSbtPlugin("org.scoverage"             % "sbt-scoverage"   % "1.9.3")
```
chuwy, line 4: SBT complains about this key:

```
arn]       lazy value coverageMinimum in object ScoverageKeys is deprecated (since v1.8.0): Use coverageMinimumStmtTotal instead
[warn]       L87:     coverageMinimum := 50,
```

Let's fix it in this commit as well.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/RegistryLookup.scala
```diff
@@ -14,25 +14,26 @@ package com.snowplowanalytics.iglu.client.resolver
...
 import scala.util.control.NonFatal
 
 // cats
-import cats.{Eval, Id}
 import cats.data.{EitherT, OptionT}
-import cats.effect.Sync
-import cats.implicits._
 import cats.effect.implicits._
+import cats.implicits._
+import cats.{Eval, Id}
```
chuwy, line 28: Can we add a commit to remove the `Eval` instance? It's been a mistake and we can fix it in this breaking release.

modules/http4s/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Http4sRegistryLookup.scala
```diff
@@ -12,31 +12,30 @@
...
+import org.typelevel.ci.CIString
 
-import com.snowplowanalytics.iglu.core.{SchemaKey, SchemaList}
-import com.snowplowanalytics.iglu.core.circe.CirceIgluCodecs._
+import scala.util.control.NonFatal
 
 object Http4sRegistryLookup {
 
-  def apply[F[_]: Sync](client: HttpClient[F]): RegistryLookup[F] =
+  def apply[F[_]: Async](client: HttpClient[F]): RegistryLookup[F] =
```
chuwy, line 31: Have you considered changing this to `F[_]: Concurrent: fs2.io.file.Files` ?

Not sure yet how big the win would be. If it means we can instantiate `Client` without `Sync`/`Async` at all - that's great and I'd stick with it. If we still need to use impure type classes for cache - it's probably not worth it (especially given that we need to implement `embeddedLookup` duplicating the original function)

chuwy (approved):
It's a bit unfortunate that CE3 turned out to be such a huge commit because of formatting changes - if possible I'd recommend to separate formatting and actual changes (fine if too much work).

Also, I strongly recommend to make it 2.0.0 release as it's clearly a very breaking change.

Otherwise - looks great 👍

modules/http4s/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Http4sRegistryLookup.scala
pondzix: Is it possible to read resources (for embedded repo) in RT manner relying only on `Files` and without `Sync`? Looking at the `Files` API you need to pass `Path` to given file and in order to do that you need to call sth like: `getClass.getResource(pathToFile)` which I believe breaks RT.

And as you mentioned we would still need to support:

* Impure implementation of `embeddedLookup` for `Id` RegistryLookup instance.
* Pure implementation of `embeddedLookup` based on `Sync` for RT RegistryLookup instance from core module. We could also replace it with the new one based on `Files`, but `Sync` would still be required for http lookup (because of scalaj-http). We would end up with sth like ```implicit def ioLookupInstance[F[_]: Sync: Files]: RegistryLookup[F]```. BTW maybe we should deprecate this instance? 

So I think it's not worth it.

Also, instead of `Async` we can use `F[_]: Concurrent: Sync` :)

chuwy: That all make sense!

chuwy (approved):

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/package.scala
```diff
@@ -27,22 +29,29 @@ import resolver.registries.Registry
...
 
   type ListLookup = Either[LookupFailureMap, SchemaList]
 
+  type CacheEntry[A] = (FiniteDuration, A)
+
+  type SchemaCacheEntry = CacheEntry[SchemaLookup]
+
+  type ListCacheEntry = CacheEntry[ListLookup]
+
+  type ListCacheKey = (String, String, Int)
```
chuwy, line 52: Can we add a small doc here what is what.

chuwy (approved):
Looks great!

I'm bit disappointed the scalafix bump causes so huge, but useless diffs. If possible I'd try to reduce them (newlines in docstrings look unusual at least) and maybe start a conversation over what the standard scalfix config should look like.

chuwy (approved):
Gave another quick look. And it looks perfect, @pondzix!

PR #203, opened 2022-07-11
Caching improvements

istreeter:
### TODOs

- [x] Consider removing `ValidatorF` as a trait, and instead directly call methods from `CirceValidator`
- [x] Revisit `checkSchemaF` -- does it need to call `V4Schema.validate(json)` ?
- [x] Use `ResolverResult` for `listSchemas` - to enable caching improvements in RDB transformer (since 1.2.0-M4)
- [x] Write lots of tests
- [ ] Port everything to the CE3 branch (master, 2.1.0 version) - this was started [here](https://github.com/snowplow/iglu-scala-client/compare/enhancement/cache-validation-by-ttl-improved-ce3) (compiles, not checked further)


snowplowcla:

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/ValidatorF.scala
```diff
@@ -0,0 +1,68 @@
...
+// circe
+import io.circe.Json
+
+/**
+ * Validates that Json conforms to a schema
+ *
+ * Unlike a simple [[Validator]], the results returned by a ValidatorF are wrapped in an effect type `F`.
+ * Implementations of ValidatorF are expected to use caches for more efficient validation.
+ */
+trait ValidatorF[F[_], K, A] {
```
pondzix, line 30: Is this trait really needed? Is it something we would like to expose publicly? I know it's like the original `Validator` but maybe we could call new methods from `CirceValidator` directly?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/ValidatorF.scala
```diff
@@ -0,0 +1,68 @@
...
+   * Get validation errors for a Json Schema
+   *
+   * Catches errors like empty `required` property or `minimum` property containing string
+   *
+   * @param key an identifier associated with the schema.  Used as a hint for when the validator
+   * may use its cache, instead of performing an expensive interpretation of the schema.
+   * @param schema JSON Schema
+   * @return list of Processing Messages with log level above warning
+   */
+  def checkSchemaF(key: Option[K], schema: Json): F[List[ValidatorError.SchemaIssue]]
```
pondzix, line 51: Could we merge this method and `validateSchemaF`? Is there any reason these 2 are separate methods (again, it's more like a question to the original `Validator`)?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
```diff
@@ -240,4 +246,65 @@ object CirceValidator extends Validator[Json] {
...
+            validateOnReadySchema(jsonschema, data).leftMap(ValidatorError.InvalidData.apply)
+          })
+
+      def checkSchemaF(
+        key: Option[(SchemaKey, Int)],
+        schema: Json
+      ): F[List[ValidatorError.SchemaIssue]] =
+        lookupOrEvaluate(key, schema)
+          .map {
+            case Right(_)      => Nil
```
pondzix, line 306: Previously, `checkSchema` called sth like `V4Schema.validate(json)`, it's not used here currently. Is it intentional?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
```diff
@@ -240,4 +246,65 @@ object CirceValidator extends Validator[Json] {
...
+                )
+                .pure[F]
+                .flatTap { result =>
+                  key.map(lookupJsonSchema.put(_, result)).sequence_
+                }
+          }
+      }
+
+      def validateF(
+        key: Option[(SchemaKey, Int)],
```
pondzix, line 291: How about we use here `Cached | NotCached` model, so we get rid of the Option? I think then we wouldn't need `key` arg, only `data` and Cached | NotCached `schema`, which would contain timestamp and schemaKey?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/ResolverCache.scala
```diff
@@ -175,11 +217,51 @@ object ResolverCache {
...
+      case (Some(Right(_)), Right(n))           => NewlyCached(n, newSeconds)
+      case (Some(Left(c)), Left(n))             => CacheFailures(c.combine(n))
+      case (Some(Right(c)), Left(_))            => AlreadyCached(c, newSeconds)
+      case (None, Left(n))                      => CacheFailures(n)
+      case (None, Right(n))                     => NewlyCached(n, newSeconds)
+      case (Some(Left(_)), Right(n))            => NewlyCached(n, newSeconds)
     }
+
+  /** The result of committing a lookup to the cache, carrying information about whether the cache was updated as a consequence */
+  private[resolver] sealed trait CacheResult[A] { self =>
```
pondzix, line 236: Is this trait needed? Do we need differentiate already/newly cached values? It looks like all we need is sth like `TimestampedItem` case class?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -130,16 +160,52 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+   *  @param value the looked-up value
+   *  @param timestamp epoch time in seconds of when the value was last cached by the resolver
+   */
+  case class Cached[A](value: A, timestamp: Int) extends ResolverResult[A]
+
+  /** The result of a lookup when the resolver is not configured to use a cache */
+  case class NotCached[A](value: A) extends ResolverResult[A]
+
+  /** The result of a lookup that ended in failure */
+  case class LookupError(value: ResolutionError) extends ResolverResult[Nothing]
```
pondzix, line 207: As we quickly map it to either in the client... maybe it's not needed? Just leave `Cached | NotCached` as `Right` and use `Left` as error?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -130,16 +160,52 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+   *
+   *  The timestamped value is helpful when the client code needs to perform an expensive
+   *  calculation derived from the looked-up value. If the timestamp has not changed since a
+   *  previous call, then the value is guaranteed to be the same as before, and the client code
+   *  does not need to re-run the expensive calculation.
+   *
+   *  @param value the looked-up value
+   *  @param timestamp epoch time in seconds of when the value was last cached by the resolver
+   */
+  case class Cached[A](value: A, timestamp: Int) extends ResolverResult[A]
```
pondzix, line 201: Can we add key here as well? And then pass whole thing to validator, without converting to `Option`?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/ResolverCache.scala
```diff
@@ -118,21 +142,28 @@ object ResolverCache {
...
     key: K
-  ): F[Option[Lookup[A]]] = {
+  ): F[Option[(Lookup[A], Int)]] = {
     val wrapped = for {
       (timestamp, lookup) <- OptionT(c.get(key))
       seconds             <- OptionT.liftF(currentSeconds[F])
       useSeconds = seconds
       if isViable(ttl, seconds, timestamp) || lookup.isLeft // isLeft
-    } yield lookup
+    } yield (lookup, timestamp)
```
pondzix, line 155: Oh, my favourite type of review comment: `Can we add some case class or type alias for that thing`? 🙂 `TimestampedItem` or sth?

BTW I recall I already added some type aliases in the latest version of client (2.x.x). Is there a plan to apply this feature to the latest version as well?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/ValidatorF.scala
istreeter: I don't have a strong reason why it should be a trait, except that I copied the original `Validator`.  I guess it exists as a trait so that you could build a `Client` around json4s or some other json library.  We have [historical support for json4s](https://github.com/snowplow/iglu-scala-core/tree/1.1.0/iglu-core-json4s/src).  On the other hand, we are now so deeply committed to circe that I don't see any call to support more than one json library.  So yeah I guess we could kill the trait and call CirceValidator methods directly.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/ValidatorF.scala
istreeter: Well yeah `validateSchemaF` is just a wrapper around `checkSchemaF`.  The point is, if you implement one then you get the other for free.  Same as, say, cats `Applicative` where if you define `ap` then you [get `map` for free](https://github.com/typelevel/cats/blob/v2.8.0/core/src/main/scala/cats/Applicative.scala#L70-L71). It's trying to be helpful.

I'm not defending it, I'm just guessing why `Validator` is as it is.

If we drop `ValidatorF` as a trait then we don't need to debate this!!!

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
istreeter: Good point, this needs more investigation.  In this implementation I am assuming that `lookupOrEvaluate` returns a `JsonSchema` only when the schema is valid against the v4 schema.  But I haven't fully checked if that's a valid assumption.  One solution could be to add a schema validation check into the `lookupOrEvaluate` function.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/ResolverCache.scala
istreeter: Yes, we definitely need to port it to the latest client!  I worked on the old branch to start with because I wanted to test it out in Enrich.  Migrating it to the CE3 branch can be a task for the next month, but let's agree on a clean API interface first.

istreeter:
Thanks @pondzix!  I implemented some suggested changes and opened a task list for the other good ideas which I've not had time for yet.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/ValidatorF.scala
pondzix: In a spirit of 'simplifying stuff': drop everything! Maybe we could have only single new (private for client)`validate` method in `CirceValidator` which would check schema (`lookupOrEvaluate`) + data? We already have `ValidationError` with `InvalidData` or `InvalidSchema` so to me it sounds ok.

pondzix: And then client would be simple: resolve schema and then validate

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -130,16 +160,37 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+   *
+   *  The timestamped value is helpful when the client code needs to perform an expensive
+   *  calculation derived from the looked-up value. If the timestamp has not changed since a
+   *  previous call, then the value is guaranteed to be the same as before, and the client code
+   *  does not need to re-run the expensive calculation.
+   *
+   *  @param value the looked-up value
+   *  @param timestamp epoch time in seconds of when the value was last cached by the resolver
+   */
+  case class Cached[A](key: SchemaKey, value: A, timestamp: Int) extends ResolverResult[A]
```
pondzix, line 189: Do we need type param `A`? We have `SchemaKey` here so we're committed to schema. And schema == `Json` value, right?

pondzix: And if it's only about schema: change name to `SchemaLookupResult`?

pondzix: Aaaand make the trait private?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
```diff
@@ -240,4 +247,71 @@ object CirceValidator extends Validator[Json] {
...
+    CreateLruMap[F, (SchemaKey, Int), Either[ValidatorError.InvalidSchema, JsonSchema]]
+      .create(100)
+      .map(validatorFImpl(_))
+
+  def validatorFImpl[F[_]: Monad](
+    lookupJsonSchema: LruMap[F, (SchemaKey, Int), Either[ValidatorError.InvalidSchema, JsonSchema]]
+  ): ValidatorF[F, Json] =
+    new ValidatorF[F, Json] {
+
+      def lookupOrEvaluate(
```
pondzix, line 263: What do you think about sth like that:

```scala
      private def lookupOrEvaluate(result: ResolverResult[Json]): F[Either[ValidatorError, JsonSchema]] = {
        result match {
          case Cached(key, schema, timestamp) =>
            evaluationCache.get((key, timestamp)).flatMap {
              case Some(alreadyEvaluatedSchema) => 
                alreadyEvaluatedSchema.pure[F]
              case None =>  
               evaluateSchema(schema)
                 .flatTap(result => evaluationCache.put((key, timestamp), result))
            }
          case NotCached(schema) =>
            evaluateSchema(schema)
        }
      }

  private def evaluateSchema[F[_] : Monad](schema: Json) = {
    Either
      .catchNonFatal(
        IgluMetaschemaFactory
          .getSchema(circeToJackson(schema), SchemaValidatorsConfig)
      )
      .leftMap(ValidatorError.schemaIssue)
      .pure[F]
  }

```
?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/ValidatorF.scala
pondzix: If we make everything private to client, make new `Client` the only public thing, we can do everything we want with internals in the future if needed, right?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
istreeter: The reason for type param `A`.... it's because `Resolver` has two main methods, `lookupSchema` and `listSchemas`.  The idea was to return `Cached[Json]` for a single lookup, and `Cached[SchemaList]` for the listing.

We should review our snowplow apps (mainly rdb transformer in shredding mode) and see if we might ever need the cache timestamp on `SchemaList`.  If we will never need a `Cached[SchemaList]` then I agree we don't need the type parameter.

>  change name to SchemaLookupResult?

Similarly to the above, I named it `ResolverResult` in case it might be useful for multiple types of lookup.  I guess we could alias it `type SchemaLookupResult = ResolverResult[Json]`.  Or I'm also happy to rename it with your suggestion if you think the lookup-flexibility is not needed.

> make the trait private?

This was deliberately public.  I was thinking for Snowplow loaders the logic would be:

- Use the resolver to lookup a `ResolverResult[Json]`
- Use a LruMap to lookup the schema-ddl column type.  For example, BigQuery loader will have a `LruMap[F, (SchemaKey, Int), Field`, to lookup [a BigQuery Field](https://github.com/snowplow/schema-ddl/blob/0.15.0/modules/core/src/main/scala/com.snowplowanalytics/iglu.schemaddl/bigquery/Field.scala)
- If it's missing from the LruMap then parse the schema and build the `Field`, which is a cpu-expensive operation.

You've already reviewed my [attempt to get this working in RDB Transformer](https://github.com/snowplow/snowplow-rdb-loader/pull/973), so you'll know I don't have it working yet as I described.  Currently in the transformer [my LruMap](https://github.com/snowplow/snowplow-rdb-loader/pull/973/files#diff-ead9dcf097f06399c326cb8072059effcf4aa4100555c876651acc0a8f7e1efcR28) is keyed by `SchemaKey`, not by `(SchemaKey, Int)`.

Our next step could be to spike solutions in RDB Transformer and BigQuery Loader, and then revisit some of these decisions here before publishing.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
istreeter: This is a definite improvement to my version!

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
pondzix: I did some tests and it seems: `IgluMetaschemaFactory.getSchema` doesn't do any validations. So you could provide invalid schema (according to V4) and you would still get JsonSchema instance. And you could validate data against it! So to preserve current behaviour of client, we still need to validate schema against V4 explicitly

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/ValidatorF.scala
pondzix: Tried this in https://github.com/snowplow/iglu-scala-client/pull/203/commits/597f025fcb60d3ac09d09efa458e7d03e96f41ef

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/CirceClient.scala
```diff
@@ -0,0 +1,62 @@
...
+import com.snowplowanalytics.iglu.core.SelfDescribingData
+import com.snowplowanalytics.lrumap.CreateLruMap
+import io.circe.{DecodingFailure, Json}
+
+final class CirceClient[F[_]] private (
+  resolver: Resolver[F],
+  schemaEvaluationCache: SchemaEvaluationCache[F]
+) {
+  def check(
+    instance: SelfDescribingData[Json] //TODO can we commit to Json here and get rid of type param?
```
istreeter, line 35: I agree, let's commit to circe Json.  I cannot think why we might one day regret this.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/CirceClient.scala
```diff
@@ -0,0 +1,62 @@
...
+  )(implicit
+    M: Monad[F], //TODO can we move those to client constructor?
```
istreeter, line 37: The pattern we use almost everywhere in Snowplow is to put implicit typeclasses on the methods, not on the constructors.

I think this was mainly Anton's influence, although I cannot remember the details of his reasoning.  I remember an occasion where a code change was made more difficult because of implicits on a constructor, and I remember Anton saying something like "I told you so" (he phrased it better)... but I cannot remember any details and I cannot find a record of the conversation.

Generally, I have never found it a problem to have implicit typeclasses on the methods, so I'd prefer to stick with what's working for us.  Unless you have a strong reason to move them?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
```diff
@@ -240,4 +248,75 @@ object CirceValidator extends Validator[Json] {
 
   private def fromValidationMessage(m: ValidationMessage): ValidatorReport =
     ValidatorReport(m.getMessage, m.getPath.some, m.getArguments.toList, m.getType.some)
+
+  private def evaluateSchema(
```
istreeter, line 252: I like how everything new is now private; you've completely minimised the api interface that we would have to maintain for future versions.  I guess this became possible after you removed the need for a public ValidatorF trait.

pondzix:
Changes for branch 1.x released in https://github.com/snowplow/iglu-scala-client/pull/208 with version 1.2.0
Changes for branch 2.x on https://github.com/snowplow/iglu-scala-client/tree/enhancement/cache-validation-by-ttl-improved-ce3

PR #206, opened 2022-09-30
Build for Scala3

hamnis:
* Fix some issues where Scala3 is stricter with types (in tests)
* Only add compiler plugins for Scala2
* Make sure we extract the valid integers correctly to jackson
* remove final from object
* Upgrade all dependencies

### DIFF
--- a/modules/core/src/main/scala/io/circe/jackson/snowplow/package.scala
+++ b/modules/core/src/main/scala/io/circe/jackson/snowplow/package.scala
@@ -45,7 +45,16 @@ package object snowplow {
           number match {
             case _: JsonBiggerDecimal | _: JsonBigDecimal =>
               number.toBigDecimal
-                .map(bigDecimal => DecimalNode.valueOf(bigDecimal.underlying))
+                .map(bigDecimal => {
+                  if (bigDecimal.isValidInt)
+                    IntNode.valueOf(bigDecimal.intValue)
+                  else if (bigDecimal.isValidLong) {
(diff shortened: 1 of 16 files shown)


modules/core/src/main/scala/io/circe/jackson/snowplow/package.scala
```diff
@@ -45,7 +45,16 @@ package object snowplow {
...
+                .map(bigDecimal => {
+                  if (bigDecimal.isValidInt)
+                    IntNode.valueOf(bigDecimal.intValue)
+                  else if (bigDecimal.isValidLong) {
+                    LongNode.valueOf(bigDecimal.longValue)
+                  } else if (bigDecimal.isWhole) {
+                    BigIntegerNode.valueOf(bigDecimal.toBigInt.underlying)
+                  } else
+                    DecimalNode.valueOf(bigDecimal.underlying)
+                })
```
istreeter, line 57: Hi @hamnis please can you help me understand why this change is needed?  Is it related to the scala 3 upgrade, or is for some other reason?  Are there any examples of when the old code would give the wrong result?

I'm happy to merge it, along with your other changes, but I like to understand the reason for things first.

hamnis: It seems like circe on scala3 sometimes parses numbers differently, and I had to do this to avoid representing numbers in jackson incrorrectly. The tests where failing without this code.

I can try to find the actual reason in circe if you want.

if you want you can comment out this block and see how the tests fail.

PR #211, opened 2022-10-25
Allow building caches based on the resolver's content - iglu-scala-client 2.x.x series version

lmath:
These changes https://github.com/snowplow/iglu-scala-client/pull/208 but off the master branch.

### DIFF
--- a/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
+++ b/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
@@ -19,6 +19,7 @@ import cats.implicits._
 import cats.{Applicative, Id, Monad}
 import com.snowplowanalytics.iglu.client.ClientError.ResolutionError
 import com.snowplowanalytics.iglu.client.resolver.registries.Registry.Get
+import com.snowplowanalytics.iglu.client.resolver.ResolverCache.TimestampedItem
 import com.snowplowanalytics.iglu.client.resolver.registries.{
   Registry,
   RegistryError,
@@ -30,7 +31,6 @@ import io.circe.{Decoder, DecodingFailure, HCursor, Json}
 
 import java.time.Instant
 import scala.collection.immutable.SortedMap
-import scala.concurrent.duration.DurationInt
 
 /** Resolves schemas from one or more Iglu schema registries */
 final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCache[F]]) {
@@ -44,25 +44,110 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
    * If any of repositories gives non-non-found error, lookup will retried
    *
    * @param schemaKey The SchemaKey uniquely identifying the schema in Iglu
+   * @return a [[Resolver.ResolverResult]] boxing the schema Json on success, or a ResolutionError on failure
+   */
+  def lookupSchemaResult(
+    schemaKey: SchemaKey
+  )(implicit
+    F: Monad[F],
+    L: RegistryLookup[F],
+    C: Clock[F]
+  ): F[Either[ResolutionError, SchemaLookupResult]] = {
+    val get: Registry => F[Either[RegistryError, Json]] = r => L.lookup(r, schemaKey)
+
+    def handleAfterFetch(
+      result: Either[LookupFailureMap, Json]
+    ): F[Either[ResolutionError, SchemaLookupResult]] =
+      cache match {
+        case Some(c) =>
+          c.putSchemaResult(schemaKey, result).map {
+            case Right(ResolverCache.TimestampedItem(i, t)) =>
+              Right(ResolverResult.Cached(schemaKey, i, t))
+            case Left(failure) => Left(resolutionError(failure))
+          }
+        case None =>
+          result
+            .bimap[ResolutionError, SchemaLookupResult](
+              resolutionError,
+              ResolverResult.NotCached(_)
+            )
+            .pure[F]
+      }
+
+    getSchemaFromCache(schemaKey).flatMap {
+      case Some(TimestampedItem(Right(schema), timestamp)) =>
+        Monad[F].pure(Right(ResolverResult.Cached(schemaKey, schema, timestamp)))
+      case Some(TimestampedItem(Left(failures), _)) =>
+        retryCached[F, Json](get, schemaKey.vendor)(failures)
+          .flatMap(handleAfterFetch)
+      case None =>
+        traverseRepos[F, Json](get, prioritize(schemaKey.vendor, allRepos.toList), Map.empty)
+          .flatMap(handleAfterFetch)
+    }
+  }
+
+  /**
+   * Tries to find the given schema in any of the provided repository refs
+   * If any of repositories gives non-non-found error, lookup will retried
+   *
+   * @param schemaKey The SchemaKey uniquely identifying the schema in Iglu
    * @return a Validation boxing either the Schema's
    *         Json on Success, or an error String
    *         on Failure
    */
   def lookupSchema(
     schemaKey: SchemaKey
-  )(implicit F: Monad[F], L: RegistryLookup[F], C: Clock[F]): F[Either[ResolutionError, Json]] = {
-    val get: Registry => F[Either[RegistryError, Json]] = r => L.lookup(r, schemaKey)
-    val resultAction = getFromCache(schemaKey).flatMap {
-      case Some(lookupResult) =>
-        lookupResult.fold(retryCached[F, Json](get, schemaKey.vendor), finish[F, Json])
-      case None =>
-        traverseRepos[F, Json](get, prioritize(schemaKey.vendor, allRepos.toList), Map.empty)
-    }
+  )(implicit
+    F: Monad[F],
+    L: RegistryLookup[F],
+    C: Clock[F]
+  ): F[Either[ResolutionError, Json]] =
+    lookupSchemaResult(schemaKey).map(_.map(_.value))
 
-    for {
-      result      <- resultAction
-      cacheResult <- cache.traverse(c => c.putSchema(schemaKey, result))
-    } yield postProcess(cacheResult.getOrElse(result))
+  /**
+   * Get list of available schemas for particular vendor and name part
+   * Server supposed to return them in proper order
+   */
+  def listSchemasResult(
+    vendor: Vendor,
+    name: Name,
+    model: Model
+  )(implicit
+    F: Monad[F],
+    L: RegistryLookup[F],
+    C: Clock[F]
+  ): F[Either[ResolutionError, SchemaListLookupResult]] = {
+    val get: Registry => F[Either[RegistryError, SchemaList]] = r => L.list(r, vendor, name, model)
+
+    def handleAfterFetch(
+      result: Either[LookupFailureMap, SchemaList]
+    ): F[Either[ResolutionError, SchemaListLookupResult]] =
+      cache match {
+        case Some(c) =>
+          c.putSchemaListResult(vendor, name, model, result).map {
+            case Right(ResolverCache.TimestampedItem(schemaList, timestamp)) =>
+              Right(ResolverResult.Cached((vendor, name, model), schemaList, timestamp))
+            case Left(failure) => Left(resolutionError(failure))
+          }
+        case None =>
+          result
+            .bimap[ResolutionError, SchemaListLookupResult](
+              resolutionError,
+              ResolverResult.NotCached(_)
+            )
+            .pure[F]
+      }
+
+    getSchemaListFromCache(vendor, name, model).flatMap {
+      case Some(TimestampedItem(Right(schemaList), timestamp)) =>
+        Monad[F].pure(Right(ResolverResult.Cached((vendor, name, model), schemaList, timestamp)))
+      case Some(TimestampedItem(Left(failures), _)) =>
+        retryCached[F, SchemaList](get, vendor)(failures)
+          .flatMap(handleAfterFetch)
+      case None =>
+        traverseRepos[F, SchemaList](get, prioritize(vendor, allRepos.toList), Map.empty)
+          .flatMap(handleAfterFetch)
+    }
   }
 
   /**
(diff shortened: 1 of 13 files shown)


snowplowcla:

modules/core/src/test/scala/com.snowplowanalytics.iglu.client/resolver/ResolverResultSpec.scala
```diff
@@ -0,0 +1,521 @@
...
+    val (state, response) = result.run(ResolverSpecHelpers.RegistryState.init).value
+
+    // Final response must not overwrite a successful one
+    val finalResult = response must beRight[SchemaLookupResult].like {
+      case ResolverResult.Cached(key, value, _) =>
+        key must beEqualTo(schemaKey) and (value must beEqualTo(Json.Null))
+    }
+
+    // Check that it attempted to get fourth schema (500 response)
+    val lookupTries = state.req must beEqualTo(4)
```
lmath, line 296: I think I'm missing something here - I can't work out why we would expect there to be a fourth request if the third request was successful and isn't stale.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/ResolverCache.scala
```diff
@@ -118,21 +156,31 @@ object ResolverCache {
       Applicative[F].pure(none)
   }
 
-  def getItem[F[_]: Monad: Clock, A, K](
+  private[resolver] def shouldCreateResolverCache(size: Int, ttl: Option[TTL]): Boolean =
```
lmath, line 159: On the 1.x branch the criteria seems to be size > 0 and if there is a TTL, then it should not be a negative number
On the 2.x branch, the criteria seems to be size >= 0 and if there is a TTL, then it should not be a negative number:
https://github.com/snowplow/iglu-scala-client/commit/d1e1a37fc05f1710ef75fd893279a8578bc8435b#diff-9e946c350a6fbf56aa3e5fe99455e9cd3d3d2dd2296d5a4a04f03bf6a060eaf2R112

but I don't think it should be possible to create a cache with cache size 0 so this change brings back the 1.x criteria

see: https://github.com/snowplow/iglu-scala-client/pull/211/files#diff-c074249fe61ae962f11d3dfc7a272de9de483bf50b281cdfd2524da3800899f1R105-R111

modules/core/src/test/scala/com.snowplowanalytics.iglu.client/resolver/ResolverResultSpec.scala
lmath: [This](https://github.com/snowplow/iglu-scala-client/pull/208/files#diff-15822dfa15d03bb4368de73705f1e90a532f64de8b958069433643649caa1a4aR242) is the same test in the 1.2.0 release

lmath: Ooops, I was convinced that the TTL was 1000 seconds but Piotr nudged me to look at the [analogous test](https://github.com/snowplow/iglu-scala-client/blob/master/modules/core/src/test/scala/com.snowplowanalytics.iglu.client/resolver/ResolverSpec.scala#L243) in `ResolverSpec` and this makes a lot more sense now.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
```diff
@@ -102,4 +106,112 @@ object CirceValidator extends Validator[Json] {
...
+
+  private def validateSchemaAgainstV4(schema: JsonNode): List[ValidatorError.SchemaIssue] = {
+    V4Schema
+      .validate(schema)
+      .asScala
+      .toList
+      .map(m => ValidatorError.SchemaIssue(m.getPath, m.getMessage))
+  }
+
+  private[client] object WithCaching {
```
voropaevp, line 129: This implementation is prone to race conditions, as LruMap updates are not atomic and need to be protected by the Semaphor. 
Check out Chris Davenport's version: [here](https://gist.github.com/ChristopherDavenport/8364fa2ca14557fce12baca6cb38d5b8#file-lrucache-scala)
 
However, those are rare and occasional item deletion is not too important in the context of this library.

voropaevp (approved):
LGTM. But fix formatting.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -44,86 +44,205 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+        traverseRepos[F, SchemaList](get, prioritize(vendor, allRepos.toList), Map.empty)
+          .flatMap(handleAfterFetch)
+    }
   }
 
+  /**
+   * Get list of available schemas for particular vendor and name part
+   * Server supposed to return them in proper order
+   */
+  def listSchemas(
```
pondzix, line 157: `listSchemasResult` uses type aliases for vendor, name and model. Can we use it here as well?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -44,86 +44,205 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
     cache match {
-      case Some(c) => c.getSchema(schemaKey)
+      case Some(c) => c.getTimestampedSchema(schemaKey)
       case None    => Monad[F].pure(None)
     }
+
+  private def getSchemaListFromCache(
+    vendor: String,
+    name: String,
+    model: Int
```
pondzix, line 196: Aaand here (type aliases)

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -44,86 +44,205 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+      case Some(c) => c.getTimestampedSchemaList(vendor, name, model)
+      case None    => Monad[F].pure(None)
+    }
+
 }
 
 /** Companion object. Lets us create a Resolver from a Json */
 object Resolver {
 
+  type SchemaListKey          = (String, String, Int) // vendor, name, model
```
pondzix, line 211: Aaand here (type aliases) :)

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -44,86 +44,205 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+     *
+     * The timestamped value is helpful when the client code needs to perform an expensive
+     * calculation derived from the looked-up value. If the timestamp has not changed since a
+     * previous call, then the value is guaranteed to be the same as before, and the client code
+     * does not need to re-run the expensive calculation.
+     *
+     * @param value     the looked-up value
+     * @param timestamp epoch time in seconds of when the value was last cached by the resolver
+     */
+    case class Cached[K, A](key: K, value: A, timestamp: TTL) extends ResolverResult[K, A]
```
pondzix, line 233: Is it more like `StorageTime` instead of `TTL`?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -211,11 +330,12 @@ object Resolver {
   def bootstrap[F[_]: Monad: InitSchemaCache: InitListCache]: F[Resolver[F]] =
     Resolver.init[F](EmbeddedSchemaCount, None, Registry.EmbeddedRegistry)
 
-  private final case class ResolverConfig(
+  final case class ResolverConfig(
     cacheSize: Int,
-    cacheTtl: Option[Int],
+    cacheTtl: Option[TTL],
```
pondzix, line 335: 👌

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/ResolverCache.scala
```diff
@@ -84,6 +108,12 @@ class ResolverCache[F[_]] private (
   ): F[Option[ListLookup]] =
     getItem(ttl, schemaLists, (vendor, name, model))
 
+  private[resolver] def getTimestampedSchemaList(vendor: String, name: String, model: Int)(implicit
```
pondzix, line 111: More type aliases (and in `putSchemaListResult`)

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/ResolverCache.scala
```diff
@@ -118,21 +156,31 @@ object ResolverCache {
...
       if isViable(ttl, currentTime, storageTime) || lookup.isLeft // isLeft
-    } yield lookup
+    } yield TimestampedItem(lookup, storageTime)
 
     wrapped.value
   }
 
+  def getItem[F[_]: Monad: Clock, A, K](
+    ttl: Option[TTL],
+    c: LruMap[F, K, (TTL, Lookup[A])],
```
pondzix, line 179: That `TTL` should be... `StorageTime`? It looks like it was `TTL` before your changes, but I think it's not correct...? Can we change it? OR can we use `CacheEntry[Lookup[A]]` for this tuple (and other tuples in this file)? Whatever you think is the easiest to read :)

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
```diff
@@ -102,4 +106,112 @@ object CirceValidator extends Validator[Json] {
...
+            case None =>
+              provideNewJsonSchema(schema)
+                .pure[F]
+                .flatTap(result => evaluationCache.put((key, timestamp), result))
+          }
+        case ResolverResult.NotCached(schema) =>
+          provideNewJsonSchema(schema).pure[F]
+      }
+    }
+    def validate2[F[_]: Monad](
```
pondzix, line 164: This one is unused?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
```diff
@@ -102,4 +106,112 @@ object CirceValidator extends Validator[Json] {
...
+  private def validateSchemaAgainstV4(schema: JsonNode): List[ValidatorError.SchemaIssue] = {
+    V4Schema
+      .validate(schema)
+      .asScala
+      .toList
+      .map(m => ValidatorError.SchemaIssue(m.getPath, m.getMessage))
+  }
+
+  private[client] object WithCaching {
+    type SchemaEvaluationKey         = (SchemaKey, TTL)
```
pondzix, line 130: Should be `StorageTime` instead of `TTL`? IIRC on 1.x there was some comment with the link to the issue, could we put something similar here as well?

pondzix (commented):
Looks great, leaving some comments mostly nitpicking about type aliases.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/ResolverCache.scala
```diff
@@ -149,12 +197,22 @@ object ResolverCache {
...
+  ): F[Either[LookupFailureMap, TimestampedItem[A]]] =
     for {
-      currentTime <- Clock[F].realTime
-      cached      <- c.get(schemaKey) // Ignore TTL invalidation
-      result = chooseResult(cached.map(_._2), freshResult)
-      _ <- c.put(schemaKey, (currentTime, result)) // Overwrite or bump TTL
+      seconds <- Clock[F].realTime
+      cached  <- c.get(schemaKey) // Ignore TTL invalidation
+      result = chooseResult(cached.map(_._2), freshResult, seconds)
+      _ <- c.put(schemaKey, (seconds, result.map(_.value))) // Overwrite or bump TTL
```
voropaevp, line 215: There are no guarantees that `c` did not change since the `get` on the previous line.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
lmath: Ah yes, you are right - that should be `StorageTime`. Fixed now, and reintroduced the comment.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
lmath: Ooooops 😳 . Thanks for spotting this; I've removed it now.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/ResolverCache.scala
lmath: Good point. I like `CacheEntry[Lookup[A]]`. Have fixed it now.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
lmath: Absolutely. I think I've got them all now.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
lmath: Agreed. Have fixed.

pondzix (approved):

CHANGELOG
```diff
@@ -1,3 +1,7 @@
+Version 2.2.0 (2022-11-01)
```
pondzix, line 1: Not a big deal, but should this date be adjusted as well? ;)

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/ResolverCache.scala
istreeter: Hey @voropaevp you are right about the race conditions.  Our problem though is we need iglu-scala-client to work with `Id` effects, not just with `IO` effects.  Which means we can't use things like `Ref` or `Semaphore` or anything that uses `Concurrent`.

In this case, any race condition is harmless, because the cache always ends up eventually in a consistent state.  But you have to convince yourself of this just by looking at the code, there is no other way.  It is not a satisfying answer.

One improvement here could be to extend our [LruMap](https://github.com/snowplow-incubator/scala-lru-map/blob/0.6.0/src/main/scala/com.snowplowanalytics.lrumap/LruMap.scala) trait to have a `update` method.  For the `Id` instances the implementation would still call `get` and `set` sequentially because that's all it can do.  But then we have a `Concurrent` instance that uses a `Semaphore`.

However I don't think that adds much value.  There is something else we could do which really would add value.

Currently, if 20 fibers call `Resolver[F].lookupSchema(...)` at the same time, then the resolver makes 20 concurrent http requests to Iglu Server (if the cache is empty).  I would love to find a way to wrap that in a Semaphore, so we only make 1 request.  Without letting the Semaphore become a performance bottleneck, of course.

lmath: Thanks for highlighting this @voropaevp. I think wrapping requests to Iglu Server in a semaphore sounds interesting @istreeter and thanks for elaborating on the need to support `Id` instances.

Re: adding an update function to scala-lru-map, interestingly there is a [get](https://www.javadoc.io/doc/com.github.ben-manes.caffeine/caffeine/latest/com.github.benmanes.caffeine/com/github/benmanes/caffeine/cache/Cache.html) function in Caffeine which is also [in Scaffeine](https://github.com/blemale/scaffeine/blob/main/src/main/scala/com/github/blemale/scaffeine/Cache.scala#L49) that seems to be designed for this pattern. 

Here's a Scastie that demos a race condition using this get-then-set pattern: 
https://scastie.scala-lang.org/HCK2VLiFSB6cKHs7CSvWMQ 
And another that uses that get function that takes an update function, showing that only one update is attempted: 
https://scastie.scala-lang.org/nAgtuOIQTpyziHpag8v4ig
(record of both scasties also in a gist [here](https://gist.github.com/lmath/17c2470b33ffeefce9832070a6174c4e))

But I'm not sure we'd want Caffeine to handle this for us, and I'm also not sure it fits into the design principles of scala-lru-map.

CHANGELOG
lmath: Aaah yes, I will fix this. Thanks.

lmath: Ooops, this should have been a PR against develop so I'll actually remove the Prepare for release... commit.

Issue #207, opened 2022-10-03
Allow building caches based on the resolver's content


pondzix:
## Current state

Client essentially consists of 2 steps:

1) Lookup schema (or list of schemas) - by resolver.
2) Validate data against resolved schema - by validator.

Some applications (e.g. enrich) use whole package - steps 1) + 2) in order to validate entities of events. There are also applications (e.g. rdb-loader or bigquery-loader) which do not need to validate data. Only step 1) - schema lookup - is the feature used by those components in order to build desired model. Usually it means converting resolved JSON schemas to proper [schema-ddl](https://github.com/snowplow/schema-ddl) model. 

## Potential improvements

It's all about caching. There is already cache for lookup in a resolver. In general, JSON schemas returned by resolver are often used by downstream components to perform some costly operation: like data validation or building schema-ddl models. And we want to cache it as well, by building different caches on top of the one from resolver.

### Timestamps in resolver 

At the moment resolver returns: 

* in schema lookup - for given schema key -> schema's content 
* in listing schemas - for given combination of vendor, name and model -> matching list of schema keys

Internally, those values before being returned by resolver are cached. It is possible to configure TTL, what means all stored items can eventually expire. If another cache is built on top of the one from resolver, we need a way to inform new cache about any changes in the original resolver's cache content. 

Returning timestamp, indicating a moment when value was cached, alongside actual content would do the job. If content in resolver's cache expires: 

* Fresh one is fetched from Iglu registry and stored with a new timestamp.
* New timestamp could trigger recalculation when used as a part of a key in downstream cache. 

### Cache for validator 

We can take advantage of changes in resolver and create correlated cache in validator. Currently, for every validation in `CirceValidator`, `Json` (circe) schema instance is converted to `JsonSchema` instance from `com.networknt` library. This operation may be expensive from CPU point of view, so let's cache it!

The structure of a new cache could look like: 

* Key - `(schemaKey, timestamp)`
* Value -  schema evaluation result `Either[ValidatorError.InvalidSchema, JsonSchema]`

The goal is also to make it backward compatible, therefore separate client and new methods in resolver, mimicking the originals but returning the new data type (with timestamp). 

See related caching changes in rdb-loader https://github.com/snowplow/snowplow-rdb-loader/issues/1086

PR #218, opened 2022-11-18
Add invalidation for the schema-list cache 2.2.0 branch (close #215)

### LINKED ISSUE #215: Add invalidation for the schema-list cache
Problem
---
Currently, schema lists are cached based on the `SchemaListKey`, which does not include the `revision` and `addition` parts of the full `SchemaKey`. 

This makes up for the breaking scenario, particularly affecting the transformer.
- New schema `1-0-1` is added to the server
- Cache still has listing for `1-*-*` as [`1-0-0`]
- Transfomer receives the event with the new version.
- Transformer would then consider the `1-0-0` to be the highest available schema for the model
- New event, which was `1-0-1` would get downcasted to the `1-0-0` oftentimes producing a bad row.

Possible solutions
--- 
- Create an API to invalidate the cache. So consumers could resolve such conflict if it arises.
- Create an API `listSimilarSchemas(k: `SchemaKey`)`. Which could detect is `k` was in the list and automatically invalidate/refresh the cache.

I think the second solution is a cleaner one.

voropaevp:


modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -111,7 +111,8 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
   def listSchemasResult(
     vendor: Vendor,
     name: Name,
-    model: Model
+    model: Model,
+    mustIncludeKey: Option[SchemaKey] = None
```
pondzix, line 115: As this is public method, such change would mean breaking compatibility? So to avoid that, we could either: 

* don't touch this method, then basically copy all that content but with `mustIncludeKey` to the new `listSchemasLike` method. 
* extract this new method body to private method, `listSchemasResult` would call it with `None`, `listSchemasLike` with `Some`

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -165,6 +170,19 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+  /**
+   * Vendor, name, model are extracted from supplied schema key to call on the `listSchemas`. The important difference
+   * from `listSchemas` is that it would invalidate cache, if returned list did not contain SchemaKey supplied in
+   * argument. Making it a safer option is latest schema bound is known.
+   */
+  def listSchemasLike(schemaKey: SchemaKey)(implicit
+    F: Monad[F],
+    L: RegistryLookup[F],
+    C: Clock[F]
+  ): F[Either[ResolutionError, SchemaList]] =
```
pondzix, line 182: Is it possible to return here full `SchemaListLookupResult`? I suspect it would be useful for caching in transformer?

modules/core/src/test/scala/com.snowplowanalytics.iglu.client/resolver/ResolverSpec.scala
```diff
@@ -404,4 +407,47 @@ class ResolverSpec extends Specification with CatsEffect {
...
+          c.some
+        }
+      )
+    )
+
+    val resultOne = resolver.listSchemasLike(schema100)
+    val resultTwo = resolver.listSchemasLike(schema101)
+
+    resultOne must beRight(SchemaList(List(schema100)))
+    resultTwo must beRight(SchemaList(List(schema100, schema101)))
```
pondzix, line 451: If there is `1-0-2` added to the iglucentral this test would fail?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
voropaevp: I have a suspicion that this method should not be public. Cached/NotCached should be transparent for the consumers.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
voropaevp: This functions in the same the the `listSchemas` does. The same argument as before as consumers shouldn't need to know if it came from the cache or not. 

The transformer's result would be cached by the iglu-client, so I won't be building another caching wrapper.

pondzix: In general yes, caching details should be hidden. But we needed that to build related cache in the transfromer -> see : https://github.com/snowplow/iglu-scala-client/issues/207 and usage in transformer https://github.com/snowplow/snowplow-rdb-loader/commit/756b4a75187f24865ce6c6db894f53b7027f9c15

That's what I meant by 'caching in transformer' - on top of the cache in iglu-client build another cache for ddl `Properties` (in case of shredded type) and cache for ddl `Field` (in case of parquet type)

pondzix: What about:
>Is it possible to return here full SchemaListLookupResult

? When we add mentioned `Field` cache to transfromer, we'd need that. So maybe it's better to add full `SchemaListLookupResult` now and map on a consumer side. Otherwise, we would have to add this method in the next release or use full `listSchemasResult(schemaKey.vendor, schemaKey.name, schemaKey.version.model, Some(schemaKey))` which is a bit verbose?

CHANGELOG
```diff
@@ -1,3 +1,7 @@
+Version 2.2.0 (2022-11-18)
```
pondzix, line 1: Aaand I think this commit shouldn't be here? ;)

voropaevp: Why not? It was created during the release in master, it should be picked to develop.

pondzix: But...2.2.0 hasn't been released yet? I think it will be when all those pending PRs are merged :)

PR #228, opened 2023-02-22
Resolver caches suffers from races and http server overload during the cold start

istreeter:


voropaevp:
I think this feature deserves a test like  such  https://github.com/snowplow/iglu-scala-client/pull/229

pondzix (approved):

PR #229, opened 2023-02-22
test for the mutex version of resolver

voropaevp:


istreeter:
Looks great thank you!  It looks like the tests fail with scala 2.12 but it works with scala 2.13.

voropaevp:
fixed

istreeter:
Merged.

PR #230, opened 2023-03-01
Schema override (1.x)

spenes:
Iglu Scala Client is updated such that it will take supersededBy into account while validating the self-describing JSONs. It should check the content of the schema first. If it contains supersededBy field, it should retrieve superseding schema.

The next change is caching. At the moment, schemas are cached in the Iglu Scala Client. With the introduction of supersededBy feature, superseding schema key should also be cached because this schema key needs to be returned to caller after validation because superseded schema key needs to be replaced with superseding schema key. Previously, cached item type was this: `Either[LookupFailureMap, Json]`. It become like this after the change: `Either[LookupFailureMap, (Json, Option[SchemaVer.Full])]`. Schema version that is stored beside schema JSON will be the version of the superseding schema.

The next change is return value of `check` function. At the moment, its return type is `Either[ClientError, Unit]`. Effectively, it returns Unit when validation is successful. However, with supersededBy feature, we want it to return superseding schema version. Then, this value will be used on the caller side.

### DIFF
--- a/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
+++ b/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
@@ -20,7 +20,7 @@ import scala.concurrent.duration.MILLISECONDS
 import cats.data._
 import cats.effect.Clock
 import cats.implicits._
-import io.circe.{Decoder, DecodingFailure, HCursor, Json}
+import io.circe.{Decoder, DecodingFailure, FailedCursor, HCursor, Json}
 import com.snowplowanalytics.iglu.core.{
   SchemaCriterion,
   SchemaKey,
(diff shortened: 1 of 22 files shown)


spenes:
Necessary changes will be made on master branch as well. I've started with 1.x since still [1.x version](https://github.com/snowplow/enrich/blob/master/project/Dependencies.scala#L65) is used in Enrich.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -63,10 +63,23 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
     L: RegistryLookup[F],
     C: Clock[F]
   ): F[Either[ResolutionError, SchemaLookupResult]] = {
-    val get: Registry => F[Either[RegistryError, Json]] = r => L.lookup(r, schemaKey)
+    val get: Registry => F[Either[RegistryError, (Json, Option[SchemaVer.Full])]] = r =>
+      EitherT(L.lookup(r, schemaKey)).semiflatMap { schema =>
+        schema.hcursor.downField("$supersededBy").as[SchemaVer.Full] match {
```
voropaevp, line 68: Consider returning an error for the malformed  `$supersededBy` value, if it exists. e.g.`match FailedCursor ...`

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/package.scala
```diff
@@ -38,7 +38,7 @@ package object resolver {
    * Json in case of success or Map of all currently failed repositories
    * in case of failure
    */
-  type SchemaLookup = Either[LookupFailureMap, Json]
+  type SchemaLookup = Either[LookupFailureMap, (Json, Option[SchemaVer.Full])]
```
voropaevp, line 41: I suggest making `(Json, Option[SchemaVer.Full])` a case class. It will make client code more readable and cleaner. As you'd give a meaningful names to the fields and would be able to provide a default constructor for None.

voropaevp (changes requested):
Looks good! Only important comment is about data structure documentation. Rest are just suggestions, that you can ignore.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
voropaevp: Should we also validate that superseded version is greater than current?

PR #233, opened 2023-03-21
Schema override

spenes:
Same changes as in [this PR](https://github.com/snowplow/iglu-scala-client/pull/230) but branched out from `master` instead [series/1.x](https://github.com/snowplow/iglu-scala-client/tree/series/1.x).

### DIFF
--- a/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/IgluCirceClient.scala
+++ b/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/IgluCirceClient.scala
@@ -18,6 +18,7 @@ import cats.effect.Clock
 import cats.implicits._
 import com.snowplowanalytics.iglu.client.resolver.registries.RegistryLookup
 import com.snowplowanalytics.iglu.client.resolver.CreateResolverCache
+import com.snowplowanalytics.iglu.client.resolver.Resolver.SupersededBy
 import com.snowplowanalytics.iglu.client.validator.CirceValidator.WithCaching.{
   InitValidatorCache,
   SchemaEvaluationCache,
@@ -43,13 +44,19 @@ final class IgluCirceClient[F[_]] private (
     M: Monad[F],
     L: RegistryLookup[F],
     C: Clock[F]
-  ): EitherT[F, ClientError, Unit] =
+  ): EitherT[F, ClientError, SupersededBy] =
     for {
-      resolverResult <- EitherT(resolver.lookupSchemaResult(instance.schema))
+      resolverResult <- EitherT(
+        resolver.lookupSchemaResult(instance.schema, resolveSupersedingSchema = true)
+      )
       validation =
         CirceValidator.WithCaching.validate(schemaEvaluationCache)(instance.data, resolverResult)
-      _ <- EitherT(validation).leftMap(_.toClientError)
-    } yield ()
+      _ <- EitherT(validation).leftMap(e =>
+        e.toClientError(resolverResult.value.supersededBy.map(_.asString))
+      )
+      // Returning superseding schema info as well since we want to inform caller that sdj is validated
+      // against superseding schema if it is superseded by another schema.
+    } yield resolverResult.value.supersededBy
 }
 
 object IgluCirceClient {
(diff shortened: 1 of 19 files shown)


modules/core/src/main/scala/com.snowplowanalytics.iglu/client/IgluCirceClient.scala
```diff
@@ -43,13 +44,17 @@ final class IgluCirceClient[F[_]] private (
     M: Monad[F],
     L: RegistryLookup[F],
     C: Clock[F]
-  ): EitherT[F, ClientError, Unit] =
+  ): EitherT[F, (ClientError, SupersededBy), SupersededBy] =
```
istreeter, line 47: Instead of returning a tuple here, did you consider amending `ClientError` to carry the extra information?  Can you share an example of the failed event that will get created by Enrich if `SupersededBy` is non-empty?

spenes: `SupersededBy` might be non-empty when there is a validation error. For that case, we might add `supersededBy` field on `ClientError.ValidationError` instead of returning it in a tuple.

modules/data/src/main/scala/com.snowplowanalytics.iglu/client/ClientError.scala
```diff
@@ -33,14 +33,17 @@ sealed trait ClientError extends Product with Serializable {
...
+
   /** Error happened during schema resolution step */
   final case class ResolutionError(value: SortedMap[String, LookupHistory]) extends ClientError {
     def isNotFound: Boolean =
       value.values.flatMap(_.errors).forall(_ == RegistryError.NotFound)
   }
 
   /** Error happened during schema/instance validation step */
-  final case class ValidationError(error: ValidatorError) extends ClientError
+  final case class ValidationError(error: ValidatorError, supersededBy: Option[String])
```
istreeter, line 45: There might be a problem here.... I _think_ this case class is used to build some of the bad row types.  And they are described by a schema [e.g. this one](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/2-0-0#L92).  So we might break bad rows if we add an extra field here.

All of that needs checking on Monday, I might be wrong.

Is there any way we can get get the relevant information in there without changing the bad rows schema?  For example, stuff the supersededBy details into ValidatorError?

spenes: I've missed that, good spot! It might not be pretty but I guess we can add a message for superseding info to `messages` list of `ValidatorError.InvalidData`. I will push the related commit shortly.

istreeter: Looking at this afresh on Monday.... maybe I was too quick to dismiss this idea to add a field to `ValidationError`.  Sure, it means we need to release a new version of the bad rows schema, but that's probably OK isn't it?  It's a non-breaking change to the schema.

spenes: Yep, I guess that would be okay. In that case, we would have to update the [schema_violations](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow.badrows/schema_violations/jsonschema/2-0-0) schema as well. Also, we need to release new version of [snowplow-badrows](https://github.com/snowplow/snowplow-badrows/blob/master/src/main/scala/com.snowplowanalytics.snowplow.badrows/Schemas.scala#L34) library with updated schema versions. It look like snowplow-badrows' latest version is on the Cats Effect 3 at the moment. We need to branch out new version from older version that doesn't contains Cats Effect 3.

I also pushed the commit that attaches the superseding info to `ValidatorError`. I know it doesn't look good because we are using the fields in `message` outside of their original purposes but superseding info is attached at least with this way. If you doesn't like this way at all, I can start to look into updating bad rows schema.

spenes: After a bit more thinking, let's do this proper way with updating bad rows schemas instead of taking short cut. Otherwise, it would be very hard to extract superseding info from bad rows. I will make necessary changes on the bad rows.

istreeter:
I have checked that this PR exactly matches [the equivalent commit that got merged in the 1.x branch](https://github.com/snowplow/iglu-scala-client/commit/4c9cd4810c67cea8733319e086ed334b4cb59eec).  So I will merge this into develop.

PR #242, opened 2023-05-18
Apps should explicitly import java.net.http.HttpClient instances of RegistryLookup

istreeter:


modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/JavaNetLookup.scala
```diff
@@ -0,0 +1,142 @@
...
+import io.circe.Json
+import io.circe.parser.parse
+
+import com.snowplowanalytics.iglu.core.circe.implicits._
+import com.snowplowanalytics.iglu.core.{SchemaKey, SchemaList}
+
+import java.net.UnknownHostException
+import scala.util.control.NonFatal
+
+object JavaNetRegistryLookup {
```
pondzix, line 28: Could we make the filename same as this `object` name? So both either `JavaNetLookup` or `JavaNewRegistryLookup`

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/RegistryLookup.scala
```diff
@@ -156,15 +90,15 @@ object RegistryLookup {
   private[registries] def toPath(prefix: String, key: SchemaKey): String =
     s"${prefix.stripSuffix("/")}/schemas/${key.toPath}"
 
-  private def toSubpath(
+  private[registries] def toSubpath(
```
pondzix, line 93: Could we move `toSubpath` methods to JavaNet... object or are these used by http4s as well?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/JavaNetLookup.scala
```diff
@@ -0,0 +1,142 @@
...
+   * @return either a `Json` on success, or `RegistryError` in case of any failure
+   *         (i.e. all exceptions should be swallowed by `RegistryError`)
+   */
+  private[registries] def httpLookup[F[_]: Sync](
+    http: Registry.HttpConnection,
+    key: SchemaKey
+  ): F[Either[RegistryError, Json]] =
+    Utils
+      .stringToUri(RegistryLookup.toPath(http.uri.toString, key))
+      .traverse(uri => Utils.getFromUri(uri, http.apikey))
```
pondzix, line 103: It looks like `Utils` contains some `java.net...` related logic, could move such stuff from `Utils` here as well?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/RegistryLookup.scala
istreeter: They are used by http4s as well.

pondzix: Hmmm so what about [this](https://github.com/snowplow/iglu-scala-client/blob/4fe1cb96e2c1a1c52a1a7f44f62237b5601cc5be/modules/http4s/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Http4sRegistryLookup.scala#L93) one? Looks like http4s has its own `toSubpath` method...?

istreeter: You were right.  I've changed it now.

pondzix (approved):

PR #245, opened 2023-05-25
Embedded and in-memory lookups should be common across http modules

istreeter:


modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/RegistryLookup.scala
```diff
@@ -80,7 +80,61 @@ object RegistryLookup {
...
+          withErrorHandling {
+            httpList(http, vendor, name, model)
+          }
+        case Registry.Embedded(_, base) =>
+          val path = toSubpath(base, vendor, name)
+          Sync[F].delay(Embedded.unsafeList(path, model))
+        case _ => Sync[F].pure(RegistryError.NotFound.asLeft)
+      }
+
+  }
```
istreeter, line 127: Hint to reviewers: start here!

This is the main new code.  Everything else is just shuffling code around, with nothing new.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Embedded.scala
```diff
@@ -0,0 +1,151 @@
...
+import cats.effect.Sync
+import cats.effect.implicits._
+import cats.implicits._
+import cats.data.EitherT
+
+// circe
+import io.circe.parser.parse
+import io.circe.Json
+
+import com.snowplowanalytics.iglu.core.SchemaList
```
oguzhanunlu, line 35: we repeatedly rename this to `SchemaKeyList` because of the confusion, maybe we can rename here until the next core release, I opened https://github.com/snowplow/iglu-scala-core/issues/100 to be included in the next core release

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Embedded.scala
```diff
@@ -0,0 +1,151 @@
...
+    try {
+      val d =
+        new File(
+          getClass.getResource(path).getPath
+        ) // this will throw NPE for missing entry in embedded repos
+      val schemaFileRegex: Regex = (".*/schemas/?" + // path to file
+        "([a-zA-Z0-9-_.]+)/" +            // Vendor
+        "([a-zA-Z0-9-_]+)/" +             // Name
+        "([a-zA-Z0-9-_]+)/" +             // Format
+        "([1-9][0-9]*)-(\\d+)-(\\d+)$").r // MODEL, REVISION and ADDITION
```
oguzhanunlu, line 91: we also repeat this regex across many places, feels like this regex should also come from the core lib (where it is tested) not to cause a potential bug

oguzhanunlu (approved):
thanks @istreeter !

PR #247, opened 2023-06-06
Http4s client should implement a retry policy (close #246)

### LINKED ISSUE #246: Http4s client should implement a retry policy
Until now, iglu-scala-client has not performed retry logic.  Even if there is a transient retriable network error, the client returns a `Left[ClientError]`.

To make things more complicated, the core library caches failed lookups. So if the surrounding application receives the `Left[ClientError]` then it cannot simply retry the call to iglu-scala-client, because it will get back the same result again.

Implementing retry logic in the "core" iglu-scala-client is tricky because we still support `Id` instances of the client, so we cannot implicitly summon a `Async[F]`.  There are ways around it, but it would be a bit hacky.

I think it's neater to put the retry logic directly into the http client.  And because we're moving towards using the http4s client for everything, I'd be happy to implement it just for that client. I reckon the retry logic should be turned on by default, because I cannot think of any Snowplow app where we would want to disable it.  We can also provide an additional constructor that allows customization of the retry settings.

### Prior work

Note, Enrich previously implemented the retry logic in the http4s client [here](https://github.com/snowplow/enrich/blob/3.8.0/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Clients.scala#LL73C6-L73C6), so in some ways this just shifting the implementation from Enrich into a different place.  However, I want to make a couple of changes compared with Enrich:

- Retry on _all_ exceptions and retriable status codes.
- Higher number of retries by default -- I suggest 3.

The Enrich implementation was there to avoid some weird "EOF" messages that we saw in failed events.  However, the fix in enrich doesn't seem to be working.  I guess it is for one of these reasons:

- The exception to catch is not really a `Command.EOF` exception. Even though we saw "EOF" in the failed event, this might be because iglu-scala-client prints the underlying root cause ([see here](https://github.com/snowplow/iglu-scala-client/blob/3.0.0/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Utils.scala#L43)), but the real exception might be a wrapper around the underlying cause.  This is why I say we should retry on all exceptions.
- One retry is not enough.  Which is why I say bump it to 3.
- Enrich is still on the CE2 branch of iglu-scala-client, which does not have the fix for issue #227.  So maybe the EOFs are happening because it makes too many concurrent requests when the cache is empty.  This can be addressed by getting Enrich onto the latest version of iglu-scala-client.

istreeter:


oguzhanunlu (approved):
thanks @istreeter !

PR #254, opened 2024-07-31
maxJsonDepth when validating JSON

istreeter:
Jira ref: PDP-1372

### DIFF
--- a/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/IgluCirceClient.scala
+++ b/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/IgluCirceClient.scala
@@ -36,8 +36,13 @@ import io.circe.{DecodingFailure, Json}
  */
 final class IgluCirceClient[F[_]] private (
   resolver: Resolver[F],
-  schemaEvaluationCache: SchemaEvaluationCache[F]
+  schemaEvaluationCache: SchemaEvaluationCache[F],
+  maxJsonDepth: Int
 ) {
+  @deprecated("Use `IgluCirceClient(resolver, cache, maxJsonDepth)`", "3.2.0")
+  def this(resolver: Resolver[F], cache: SchemaEvaluationCache[F]) =
+    this(resolver, cache, Int.MaxValue)
+
   def check(
     instance: SelfDescribingData[Json]
   )(implicit
@@ -50,7 +55,8 @@ final class IgluCirceClient[F[_]] private (
         resolver.lookupSchemaResult(instance.schema, resolveSupersedingSchema = true)
       )
       validation =
-        CirceValidator.WithCaching.validate(schemaEvaluationCache)(instance.data, resolverResult)
+        CirceValidator.WithCaching
+          .validate(schemaEvaluationCache)(instance.data, resolverResult, maxJsonDepth)
       _ <- EitherT(validation).leftMap(e =>
         e.toClientError(resolverResult.value.supersededBy.map(_.asString))
       )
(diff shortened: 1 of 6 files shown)


modules/core/src/main/scala/io/circe/jackson/snowplow/package.scala
```diff
@@ -34,7 +34,7 @@ package object snowplow {
    * @param json instance of circe's Json
    * @return converted JsonNode
    */
-  final def circeToJackson(json: Json): JsonNode =
+  final def circeToJackson(json: Json, maxDepth: Int): JsonNode =
```
istreeter, line 37: **TODO**: Currently `maxDepth` is decremented on each recursion, but it is otherwise ignroed.  Change this function so it returns an error if `maxDepth` decreased below zero.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/IgluCirceClient.scala
```diff
@@ -61,21 +63,36 @@ final class IgluCirceClient[F[_]] private (
 
 object IgluCirceClient {
 
+  @deprecated("Use `parseDefault(json, maxJsonDepth)`", "3.2.0")
   def parseDefault[F[_]: Monad: CreateResolverCache: InitValidatorCache](
     json: Json
+  ): EitherT[F, DecodingFailure, IgluCirceClient[F]] =
+    parseDefault(json, Int.MaxValue)
```
istreeter, line 70: `MaxValue` for backwards compatibility.  This method is now deprecated anyway.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/IgluCirceClient.scala
```diff
@@ -61,21 +63,36 @@ final class IgluCirceClient[F[_]] private (
...
-      client   <- EitherT.liftF(fromResolver(resolver, config.cacheSize))
+      client   <- EitherT.liftF(fromResolver(resolver, config.cacheSize, maxJsonDepth))
     } yield client
 
+  @deprecated("Use `fromResolver(resolver, cacheSize, maxJsonDepth)`", "3.2.0")
   def fromResolver[F[_]: Monad: InitValidatorCache](
     resolver: Resolver[F],
     cacheSize: Int
+  ): F[IgluCirceClient[F]] =
+    fromResolver(resolver, cacheSize, Int.MaxValue)
```
istreeter, line 87: MaxValue for backwards compatibility.  This method is now deprecated anyway.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
```diff
@@ -88,25 +88,26 @@ object CirceValidator extends Validator[Json] {
     V4SchemaInstance.getSchema(new ObjectMapper().readTree(MetaSchemas.JsonSchemaV4Text))
 
   def validate(data: Json, schema: Json): Either[ValidatorError, Unit] = {
-    val jacksonJson = circeToJackson(schema)
+    val jacksonJson = circeToJackson(schema, Int.MaxValue)
```
istreeter, line 91: MaxValue for backwards compatibility.

This method is only ever used by `Client`, and I have plans to deprecate the old Client.  In future we should support only `IgluCirceClient` which is newer and better.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
```diff
@@ -88,25 +88,30 @@ object CirceValidator extends Validator[Json] {
...
-        validateOnReadySchema(schema, data).leftMap(ValidatorError.InvalidData.apply)
+        validateOnReadySchema(schema, data, Int.MaxValue).leftMap(ValidatorError.InvalidData.apply)
       }
   }
 
-  def checkSchema(schema: Json): List[ValidatorError.SchemaIssue] = {
-    val jacksonJson = circeToJackson(schema)
+  @deprecated("Use `checkSchema(schema, maxJsonDepth)`", "3.2.0")
+  def checkSchema(schema: Json): List[ValidatorError.SchemaIssue] =
+    checkSchema(schema, Int.MaxValue)
```
istreeter, line 100: This method is used by:

- `Client` (via `Validator`) which I think we can deprecate
- igluctl, which could switchover to use `checkSchema(schema, maxJsonDepth)`

istreeter (commented):
@spenes I can't approve this because I contributed half the PR.  But I'm happy with the half the you contributed.  So if you're happy with my half then I reckon we can go ahead and merge it.

PR #255, opened 2024-08-07
[WIP] new function not based on list endpoint

benjben:
- [ ] use schemas from different repos in tests


modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -181,6 +181,93 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
   ): F[Either[ResolutionError, Json]] =
     lookupSchemaResult(schemaKey).map(_.map(_.value.schema))
 
+  def lookupSchemasUntil(
+    maxSchemaKey: SchemaKey
+  )(implicit
+    F: Monad[F],
+    L: RegistryLookup[F],
+    C: Clock[F]
+  ): F[Either[ResolutionError, List[Json]]] = {
```
benjben, line 190: Hi @istreeter , `ResolutionError` doesn't tell for which schema the error is. I don't know much about how this function will get called, should I return e.g. `case class SchemaResolutionError(schema: SchemaKey, error: ResolutionError)` instead ?

istreeter: Good question.  Ultimately the return type is used build a "loader iglu error" flavour of bad row.  [Here are the sub-types](https://github.com/snowplow/snowplow-badrows/blob/2.3.0/src/main/scala/com.snowplowanalytics.snowplow.badrows/FailureDetails.scala#L169-L182) we can pick from.

Previously, we have been turning it into a `SchemaListNotFound` bad row:

```
/** Loader couldn't fetch a `SchemaList` - entity necessary for proper shredding with migrations */
final case class SchemaListNotFound(schemaCriterion: SchemaCriterion, error: ClientError) extends LoaderIgluError
```

But after this PR I guess it's more appropriate to return a `IgluError`:

```
/** "Classic" error - loader couldn't find a schema (`ValidationError` should never happen) */
final case class IgluError(schemaKey: SchemaKey, error: ClientError) extends LoaderIgluError
```

So it would be great if your function can return sufficient information that we can construct that failure class.

benjben: I created [SchemaResolutionError](https://github.com/snowplow/iglu-scala-client/blob/not_use_list/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala#L487) :

```
case class SchemaResolutionError(schemaKey: SchemaKey, error: ResolutionError)
```
which is more specific than
```
final case class IgluError(schemaKey: SchemaKey, error: ClientError) extends LoaderIgluError
```
and can easily be used to construct it.

istreeter:
Since you said

> don't hesitate to propose an alternative

I came up with [this](https://github.com/snowplow/iglu-scala-client/blob/ef2c67104bd9ff66ce4b63f627958c9cc4b8b8b3/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala#L184-L215).  The tests don't all pass, so I've probably got something wrong.  But hopefully you'll see what it's trying to do.

benjben:
Thanks @istreeter ! Your implementation is much more simple, I'll go with it !
@pondzix I think that you are reviewing the PR, bear with me while I do the change, I also need to change the return type of the function for `common-streams` and I need to add a test that should have failed with Ian's change (they all pass).

pondzix:
@istreeter please, we don't use `go` here 😁

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -181,6 +181,101 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+        finished: Boolean
+      )
+      Monad[F]
+        .iterateWhileM[Fetched](
+          Fetched(None, List.empty[Json], false)
+        ) { case Fetched(error, schemas, finished) =>
+          val nextSchemaKey =
+            schemaKey.copy(version = schemaKey.version.copy(addition = schemas.size))
+          lookupSchema(nextSchemaKey).map {
+            case Left(resolutionError) if resolutionError.isNotFound && schemas.nonEmpty =>
```
istreeter, line 210: This line is causing me much anxiety.  The return value of `isNotFound` is only true if **every** Iglu repo returns a 404.  If any single repo returns, 500 then here you return something which causes a failed event.

In our typical setups, we have 1 private iglu server, plus up to 4 mirrors of Iglu Central.  We use 4 mirrors to provide redundancy if any single mirror becomes unreachable.

So imagine a situation where:

- The private repo contains `1-0-0` and `1-1-0` of a schema
- The schema does not exist in Iglu Central, so we always get 404 from Iglu Central
- One of the Iglu Central mirrors has become unhealthy, so it always returns 500

Here are the response codes for each lookup:

| Version | Private Iglu repo | Iglu Central | Iglu Central mirror 1 | Iglu Central mirror 2 |
|---------|-------------------|--------------|-----------------------|-----------------------|
| 1-0-0   | 200               | 404          | 404                   | 500                   |
| 1-0-1   | 404               | 404          | 404                   | 500                   |
| 1-1-0   | 200               | 404          | 404                   | 500                   |

In this scenario, ideally the result should successfully return the `1-0-0` and `1-1-0` schemas.  But your current implementation returns a schema resolution error in this scenario.

I'm still trying to work out what is the best behaviour in this scenario.  Do you have any ideas?

My doubts still are relevant even if you switch to the alternative implementation.  We need to decide whether to check `isNotFound`.

pondzix: What if one of those unhealthy repositories contains `1-0-1`? We never know

pondzix: Interesting idea! Now vendor prefix is only used to prioritize repos, right? We could go even further and don't do lookup for repos without a vendor match (assuming there is a match somewhere else). Like, "don't even try, there is no chance to find anything there" with given schema vendor.

benjben: > What if one of those unhealthy repositories contains 1-0-1? We never know
 
But we do know that Iglu central and all its mirrors should contain the same schemas, so if one returns 404, they all should.

benjben: Ideally, we would want that if Iglu Central or one of its mirror returns `404` for a schema, we ignore the `5xx` of the others (not including private repo of course). 

This is what the resolvers looks like for customers.

On GCP:
```
{
  "schema": "iglu:com.snowplowanalytics.iglu/resolver-config/jsonschema/1-0-2",

  "data": {
    "cacheSize": 500,
    "cacheTtl": 3600,
    "repositories": [
      {
        "name": "Iglu Central",
        "priority": 10,
        "vendorPrefixes": [ ],
        "connection": {
          "http": {
            "uri": "http://iglucentral.com"
          }
        }
      },
      {
        "name": "Iglu Central - GCP Mirror",
        "priority": 10,
        "vendorPrefixes": [ ],
        "connection": {
          "http": {
            "uri": "http://mirror01.iglucentral.com"
          }
        }
      },
      {
       ... private repo ...
      }
    ]
  }
}
```

On AWS:
```
{
  "schema": "iglu:com.snowplowanalytics.iglu/resolver-config/jsonschema/1-0-2",
  "data": {
    "cacheSize": 500,
    "cacheTtl": 600,
    "repositories": [
      {
        "name": "Iglu Central",
        "priority": 1,
        "vendorPrefixes": [],
        "connection": {
          "http": {
            "uri": "http://iglucentral.com"
          }
        }
      },
      {
        "name": "Iglu Central - Mirror 01",
        "priority": 1,
        "vendorPrefixes": [],
        "connection": {
          "http": {
            "uri": "http://mirror01.iglucentral.com"
          }
        }
      },
      {
        "name": "Iglu Central - Server Mirror eu1",
        "priority": 10,
        "vendorPrefixes": [],
        "connection": {
          "http": {
            "uri": "https://com-iglucentral-......iglu.snplow.net/api"
          }
        }
      },
      {
        "name": "Iglu Central - Server Mirror eu2",
        "priority": 10,
        "vendorPrefixes": [],
        "connection": {
          "http": {
            "uri": "https://com-iglucentral-......iglu.snplow.net/api"
          }
        }
      },
      {
      ... private repo ...
      }
    ]
  }
}
```

On Azure:
```
{
  "schema": "iglu:com.snowplowanalytics.iglu/resolver-config/jsonschema/1-0-2",
  "data": {
    "cacheSize": 500,
    "cacheTtl": 600,
    "repositories": [
      {
        "name": "Iglu Central",
        "priority": 1,
        "vendorPrefixes": [],
        "connection": {
          "http": {
            "uri": "http://iglucentral.com"
          }
        }
      },
      {
        "name": "Iglu Central - Mirror 01",
        "priority": 1,
        "vendorPrefixes": [],
        "connection": {
          "http": {
            "uri": "http://mirror01.iglucentral.com"
          }
        }
      },
      {
        "name": "Iglu Central - Server Mirror eu1",
        "priority": 10,
        "vendorPrefixes": [],
        "connection": {
          "http": {
            "uri": "https://com-iglucentral-.....iglu.snplow.net/api"
          }
        }
      },
      {
        "name": "Iglu Central - Server Mirror eu2",
        "priority": 10,
        "vendorPrefixes": [],
        "connection": {
          "http": {
            "uri": "https://com-iglucentral-......iglu.snplow.net/api"
          }
        }
      },
      {
         ... private repo ...
      },
      {
       ... private repo ...
      }
    ]
  }
}
```

In all cases the name contains `Iglu Central` and the `uri` `iglucentral`. We could use that to judge if a repo is Iglu Central or not.

I'm very aware that this is not ideal but in my opinion this is better than returning failed events if a mirror is unhealthy.

The ideal way would be to update the config with a field that says if a repo is Iglu Central or a mirror.

WDYT ?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -181,6 +188,58 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+    // it should be considered NotFound, even if one of them returned an error
+    def isNotFound(error: ResolutionError): Boolean = {
+      val (igluCentral, custom) = error.value.partition { case (repo, _) =>
+        allIgluCentral.contains(repo)
+      }
+      custom.values.flatMap(_.errors).forall(_ == RegistryError.NotFound) &&
+      (igluCentral.isEmpty || igluCentral.values.exists(
+        _.errors.forall(_ == RegistryError.NotFound)
+      ))
+    }
```
benjben, line 209: Hi @istreeter , what do you think about this implementation ? `isNotFound` is usually [defined](https://github.com/snowplow/iglu-scala-client/blob/master/modules/data/src/main/scala/com.snowplowanalytics.iglu/client/ClientError.scala#L40-L41) in `ClientError`, but there it has only access to `SortedMap[String, LookupHistory]` and it can't infer which repos are for Iglu Central, only the `Resolver`, which has the list of all `Registry`s, can ([here](https://github.com/snowplow/iglu-scala-client/blob/not_use_list/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala#L41)). Changing that now would be quite a big redesign and I don't think that it should happen in this ticket. I still need to add a unit test for this function.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -38,6 +38,13 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
   private[client] val allRepos: NonEmptyList[Registry] =
     NonEmptyList[Registry](Registry.EmbeddedRegistry, repos)
 
+  private val allIgluCentral = repos.collect {
+    case Registry.Http(config, connection)
+        if config.name.toLowerCase.matches(".*iglu[- ]?central.*") && connection.uri.toString
+          .contains("iglucentral") =>
```
istreeter, line 44: I think it looks fragile to match on the name.  Because this constrains us to be very careful how we name things in config files.  I think that could very easily break if someone does a well-meaning tidy-up of a config file.

But I do like the match on the uri, because that is a more controlled string.  I guess you could match on `uri.host` instead of `uri.toString`?  And maybe match word boundaries with a regex?

```scala
uri.host.matches(""".*\biglucentral\b.*""")
```

For any other people reading this -- we discussed a lot in a private channel about matching on `iglucentral` in a host name.  We all agree it's not ideal, but we're going to do it anyway because it's the only way to treat the mirrors as a special case, unless we do a bigger refactor.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -380,6 +439,8 @@ object Resolver {
   type SchemaListLookupResult = ResolverResult[SchemaListKey, SchemaList]
   type SupersededBy           = Option[SchemaVer.Full]
 
+  case class RawSchema(schemaKey: SchemaKey, json: Json)
```
istreeter, line 442: This case class is quite similar to a `SelfDescribingSchema[Json]`.

I think I need to see how this is used in common-streams or RDB Loader before I can comment on the return type.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -181,6 +188,58 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+    C: Clock[F]
+  ): F[Either[SchemaResolutionError, List[RawSchema]]] = {
+
+    // If Iglu Central or any of its mirrors doesn't have a schema,
+    // it should be considered NotFound, even if one of them returned an error
+    def isNotFound(error: ResolutionError): Boolean = {
+      val (igluCentral, custom) = error.value.partition { case (repo, _) =>
+        allIgluCentral.contains(repo)
+      }
+      custom.values.flatMap(_.errors).forall(_ == RegistryError.NotFound) &&
```
istreeter, line 205: This line will always return true if `custom` is an empty list.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -38,6 +38,13 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
   private[client] val allRepos: NonEmptyList[Registry] =
     NonEmptyList[Registry](Registry.EmbeddedRegistry, repos)
 
+  private val allIgluCentral = repos.collect {
```
istreeter, line 41: It would be nice to have a type annotation on this line.

istreeter (commented):
Would be nice to see some scaladoc on the public-facing methods.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
benjben: > I think it looks fragile to match on the name. Because this constrains us to be very careful how we name things in config files. I think that could very easily break if someone does a well-meaning tidy-up of a config file.

I wanted to put as much criteria as I could in the filter, but I agree that this is not very future-proof and that the URI is probably enough.

> I guess you could match on uri.host instead of uri.toString

Good idea !

> And maybe match word boundaries with a regex?

It's not as future-proof as without the boundaries, but seems not too dangerous.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
benjben: > This case class is quite similar to a SelfDescribingSchema[Json].

It's similar but there is a big difference : for `SelfDescribingSchema[Json]` the parsing of the `Json` has already happened successfully.

See how it is used in `common-streams` : https://github.com/snowplow-incubator/common-streams/blob/main/modules/loaders-common/src/main/scala/com/snowplowanalytics/snowplow/loaders/transform/SchemaProvider.scala#L31

I considered returning `SelfDescribingSchema[Json]` instead of `Json` but I wanted to leave the parsing where it currently is. I'm happy to change it if you think that it's better. I'm not very familiar with all the loaders that will cal `lookupSchemasUntil` and what the change involves.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
benjben: :facepalm: Good spot !

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
istreeter: `SelfDescribingSchema[Json]` is the schema before parsing.  And `SelfDescribingSchema[Schema]` is the schema after parsing.  Does that change your answer at all?

benjben: Oh sorry I missed that ! Then it's exactly what I need, perfect !

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -181,6 +187,69 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+   * If Iglu Central or any of its mirrors doesn't have a schema,
+   * it should be considered NotFound, even if one of them returned an error.
+   */
+  private[resolver] def isNotFound(error: ResolutionError): Boolean = {
+    val (igluCentral, custom) = error.value.partition { case (repo, _) =>
+      allIgluCentral.contains(repo)
+    }
+    (igluCentral.isEmpty || igluCentral.values.exists(
+      _.errors.forall(_ == RegistryError.NotFound)
+    )) && custom.values.flatMap(_.errors).forall(_ == RegistryError.NotFound)
```
istreeter, line 200: I _think_ the `.forall`s should be `.exists`.  Because if a registry returns an error and then later a NotFound, then we can treat that as NotFound and ignore the earlier error.  See what you think.....

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -38,6 +38,12 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
   private[client] val allRepos: NonEmptyList[Registry] =
     NonEmptyList[Registry](Registry.EmbeddedRegistry, repos)
 
+  private val allIgluCentral: Set[String] = repos.collect {
+    case Registry.Http(config, connection)
+        if Option(connection).filter(_.uri.getHost.matches(""".*\biglucentral\b.*""")).isDefined =>
```
istreeter, line 43: Normally I've seen `Option(something)` used whenever there is chance the `something` might be a null.  But here I'd be surprised if `connection` is null, so I cannot work out why we need to wrap the connection in an Option.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -181,6 +187,69 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+              SelfDescribingSchema(SchemaMap(currentSchemaKey), json) :: acc
+            )
+          else
+            Monad[F].pure(
+              Right((SelfDescribingSchema(SchemaMap(currentSchemaKey), json) :: acc).reverse)
+            )
+      }
+    }
+
+    go(SchemaVer.Full(maxSchemaKey.version.model, 0, 0), Nil)
```
benjben, line 250: Hi @istreeter ,
I now want to return a `NonEmptyVector`. I see two approaches:
1. Manually fetch `model-0-0` and then call `go()` with a `NonEmptyVector` as `acc`
2. Keep the existing and map the result like
```
  case head :: tail => create non empty vector
  case _ => create SchemaResolutionError, although we know that it's not possible
```
I don't like both approaches, but I'd still pick 2 as I don't like to need some action before calling `go()`. WDYT ?

istreeter: Sounds like 1 is the correct way to do it.  Unless that becomes horrifically ugly.  But I'm hopeful there will be a way to do 1 without ruining the simplicity of the code.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -181,6 +187,82 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+      .leftMap(SchemaResolutionError(firstSchemaKey, _))
+      .flatMap { firstJson =>
+        val acc = NonEmptyList.one(SelfDescribingSchema(SchemaMap(firstSchemaKey), firstJson))
+
+        if (maxSchemaKey.version.addition == 0 && maxSchemaKey.version.revision == 0)
+          EitherT.rightT[F, SchemaResolutionError](acc)
+        else
+          EitherT(go(SchemaVer.Full(maxSchemaKey.version.model, 0, 1), acc))
+      }
+      .value
```
benjben, line 263: I tried with a for loop but I thought that it looks ~better~ less worse like that

istreeter: How about doing it like [this](https://github.com/snowplow/iglu-scala-client/commit/a6a8616c5f216f0367be3f673159769f4fb2b9a9)?

benjben: I never considered that `go` could have a different return type than the function :facepalm:

istreeter (approved):

PR #258, opened 2024-09-26
Add caching to lookupSchemasUntil function

spenes:
Jira ref: PDP-1479

This PR adds caching to `lookupSchemasUntil` function. Also, `lookupSchemasUntilResult` function is implemented as well. It is the variant of `lookupSchemasUntil` that returns `ResolverResult` which contains information about cahe item.

### DIFF
--- a/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
+++ b/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
@@ -201,6 +201,19 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
   }
 
   /**
+   * The variant of lookupSchemasUntilResult that returns the result
+   * that isn't wrapped with ResolverResult
+   */
+  def lookupSchemasUntil(
+    maxSchemaKey: SchemaKey
+  )(implicit
+    F: Monad[F],
+    L: RegistryLookup[F],
+    C: Clock[F]
+  ): F[Either[SchemaResolutionError, NonEmptyList[SelfDescribingSchema[Json]]]] =
+    lookupSchemasUntilResult(maxSchemaKey).map(_.map(_.value))
+
+  /**
    * Looks up all the schemas with the same model until `maxSchemaKey`.
    * For the schemas of previous revisions, it starts with addition = 0
    * and increments it until a NotFound.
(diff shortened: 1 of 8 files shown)


project/BuildSettings.scala
```diff
@@ -68,7 +68,7 @@ object BuildSettings {
   // clear-out mimaBinaryIssueFilters and mimaPreviousVersions.
   // Otherwise, add previous version to set without
   // removing other versions.
-  val mimaPreviousVersions = Set("3.0.0")
+  val mimaPreviousVersions = Set()
```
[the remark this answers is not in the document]
istreeter: Fine by me if we make it 4.0.0

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -209,49 +222,95 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+            } yield fixed
+        }
+      }
+
+    getSchemaContentListFromCache(maxSchemaKey).flatMap {
+      case Some(TimestampedItem(Right(i), t)) =>
+        Monad[F].pure(Right(ResolverResult.Cached(maxSchemaKey, i, t)))
+      case Some(TimestampedItem(Left(_), _)) | None =>
+        lockAndLookup
+    }
```
benjben, line 306: On line 305 we know that the first call to `getSchemaContentListFromCache` returned either `Left` or `None`.
Yet on line 290 after we call it a second time we check again if it is successful or not.
Is it because there might be several parallel calls to the function and the cache can get updated between the 2 calls ?

modules/core/src/test/scala/com.snowplowanalytics.iglu.client/resolver/ResolverResultSpec.scala
```diff
@@ -821,4 +830,265 @@ class ResolverResultSpec extends Specification with ValidatedMatchers with CatsE
...
+      checkResult(notCached214, response214, schemaKey214) and
+      checkResult(notCached103_1, response103_1, schemaKey103) and
+      checkResult(notCached103_2, response103_2, schemaKey103) and
+      (cached100 mustEqual notCached100_1) and
+      (cached102 mustEqual notCached102) and
+      (cached200 mustEqual notCached200) and
+      (cached214 mustEqual notCached214) and
+      (cached103 mustEqual notCached103_1)
+
+    result.run(ResolverSpecHelpers.RegistryState.init).value._2
```
benjben, line 1092: TBH I didn't dig too much into the test but it seems to be great testing !

benjben (approved):
:ok_hand:

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
spenes: Exactly!

PR #260, opened 2024-11-15
PDP-1562 Consider all `ClientFailure`s as NotFound in `lookupSchemasUntil`

benjben:


modules/http4s/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Http4sRegistryLookup.scala
```diff
@@ -126,7 +126,7 @@ object Http4sRegistryLookup {
           .handleError { e =>
             RegistryError.ClientFailure(s"Could not decode server response. $e").asLeft[A]
           }
-      case Status.ClientError(response) if response.status.code == 404 =>
+      case Status.ClientError(response) if response.status.code / 100 == 4 =>
```
istreeter, line 129: You are already matching on `Status.ClientError(response)` and that will match anything with response code 4xx.  So you don't need the additional `if response.status.code / 100 == 4`.  This can be simply

```scala
case Status.ClientError(response) =>
```

[Here is](https://github.com/http4s/http4s/blob/v0.23.29/core/shared/src/main/scala/org/http4s/Status.scala#L100-L115) the relevant code from http4s.

istreeter: Well it makes the code more complicated and I don't (yet) see a reason.  So far the only reason for maybe doing this was:

> scenarios where we could be missing information for the troubleshooting. For instance, in case of wrong API key

....but what I'm trying to say is that I don't think we have that troubleshooting ability at the moment anyway.  Because Iglu Server always returns 404.  So we already cannot use the BadRow as a way to see if an API key was wrong.

Are there any other scenarios where it might be helpful to make you suggested change?

benjben: > Are there any other scenarios where it might be helpful to make you suggested change?

I have no idea... I'm trying to find good reasons to not do the change but I just don't know if some bad rows with `"Unexpected server response: $body"` are ever used for troubleshooting. I find it a bit scary to remove information like that but after learning that even a wrong API key is treated as a 404, I'm less scared. I updated.

istreeter (commented):
This looks good, but we need the same thing again [on these registry lookups](https://github.com/snowplow/iglu-scala-client/blob/master/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/JavaNetRegistryLookup.scala) which is used in the batch transfomer.

istreeter (approved):

PR #262, opened 2025-02-20
Remove entropy from client failures

oguzhanunlu:
This PR removes details from client failures and include status code only so that entropy is minimized.

ref: PDP-1642

### DIFF
--- a/modules/http4s/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Http4sRegistryLookup.scala
+++ b/modules/http4s/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Http4sRegistryLookup.scala
@@ -134,10 +134,8 @@ object Http4sRegistryLookup {
           RegistryError.RepoFailure(error).asLeft
         }
       case Status.ClientError(response) =>
-        response.bodyText.compile.string.map { body =>
-          val error = s"Unexpected server response: $body"
-          RegistryError.ClientFailure(error).asLeft
-        }
+        val error = s"Unexpected server response: ${response.status.code}"
+        (RegistryError.ClientFailure(error): RegistryError).asLeft[A].pure[F]
       case response =>
         response.bodyText.compile.string.map { body =>
           val error = s"Unexpected response: $body"
--- a/modules/http4s/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Http4sRegistryLookup.scala
+++ b/modules/http4s/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Http4sRegistryLookup.scala
@@ -134,7 +134,7 @@ object Http4sRegistryLookup {
           RegistryError.RepoFailure(error).asLeft
         }
       case Status.ClientError(response) =>
-        val error = s"Unexpected server response: ${response.status.code}"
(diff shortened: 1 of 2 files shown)


modules/http4s/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/Http4sRegistryLookup.scala
```diff
@@ -134,10 +134,8 @@ object Http4sRegistryLookup {
           RegistryError.RepoFailure(error).asLeft
         }
       case Status.ClientError(response) =>
-        response.bodyText.compile.string.map { body =>
-          val error = s"Unexpected server response: $body"
-          RegistryError.ClientFailure(error).asLeft
-        }
+        val error = s"Unexpected server response: ${response.status.code}"
```
istreeter, line 137: Maybe "Unexpected response code" is a better message?

oguzhanunlu: definitely, not every response comes from server

modules/http4s/src/test/scala/com.snowplowanalytics.iglu/client/resolver/registries/Http4sRegistryLookupSpec.scala
```diff
@@ -84,6 +86,44 @@ class Http4sRegistryLookupSpec extends Specification with CatsEffect {
...
+        )
+      val schemaKey = SchemaKey("com.myvendor", "status", "jsonschema", SchemaVer.Full(42, 42, 42))
+
+      val client = HttpClient[IO] { _ =>
+        val dsl = new Http4sDsl[IO] {}; import dsl._
+        Resource.eval(Forbidden("forbidden"))
+      }
+
+      Http4sRegistryLookup(client).lookup(repositoryRef, schemaKey).map { result =>
+        result should beLeft(ClientFailure(s"Unexpected server response: ${Forbidden.code}"))
```
istreeter: I think you should assert that the string matches the exact string you expect, i.e. check for

```
"Unexpected server response: 403"
```

In my experience, in tests it is unhelpful to be too clever with using stringified objects.

Imagine http4s makes a code change in a future release, where they change the `toString` method on `Forbidden`:

```scala
object Forbidden extends Status {
  override def toString(): String = "foo"
}
```

Would your unit test pass or fail if http4s made that change?

I know that sounds like a silly example, but I can remember bugs/incidents caused by exactly this mistake!  There was a bug in the old collector, where akka changed how they serialize sensitive headers.  We didn't notice that akka had changed their implementation, because our unit tests made the same mistake as the code under test.

istreeter: Found the collector bug I'm referring to: snowplow/stream-collector#210

oguzhanunlu: I see what you mean, passing tests would trick us if/when toString changes, updating

istreeter (approved):
Nice :)

PR #265, opened 2025-10-24
Upgrade json-schema-validator from 1.0.76 to 1.5.8

oguzhanunlu:
ref: https://snplow.atlassian.net/browse/PDP-2203

# Upgrade json-schema-validator from 1.0.76 to 1.5.8

  ## Context

  The Self-Serve team needs to validate user-submitted configs against schemas in Iglu Central. This functionality exists in Iglu Scala Client but cannot be used due to a dependency conflict.

  Console uses `com.networknt.json-schema-validator` version **1.5.8** (released 2024-06-27), while Iglu Scala Client uses version **1.0.76** (released 2022-12-19).

  This PR upgrades Iglu Client to use the newer version, enabling the Self-Serve team to leverage existing schema fetching and validation functionality.

  ## Changes

  ### 1. Dependency Updates

  | Dependency | Old Version | New Version | Reason |
  |------------|-------------|-------------|--------|
  | `json-schema-validator` | 1.0.76 | 1.5.8 | Main upgrade to resolve conflict |
  | `jackson-databind` | 2.14.1 | 2.18.3 | Compatibility requirement for validator 1.5.8 |
  | `commons-lang3` | (transitive) | 3.17.0 (explicit) | Made explicit for better dependency management |
  | `sbt` | 1.8.2 | 1.11.7 | Required for Sonatype Central Portal migration |

  ### 2. Code Changes in `CirceValidator`

  The major version jump required handling several breaking API changes:

  #### Schema Loading API
  - **Old (1.0.76)**: `URIFetcher` with `fetch(URI): InputStream`
  - **New (1.5.8)**: `SchemaLoader` with `getSchema(AbsoluteIri): InputStreamSource`

  #### JsonMetaSchema Builder
  - **Old**: `.addKeyword()`, `.addMetaSchema()`, `.uriFetcher()`
  - **New**: `.keyword()`, `.metaSchema()`, `.schemaLoaders()`

  #### Schema Factory Configuration
  - Removed deprecated methods: `.forceHttps()`, `.removeEmptyFragmentSuffix()`
  - Migrated `SchemaValidatorsConfig` to builder pattern

  #### ValidationMessage API
  - **Old**: `m.getPath` (returns String), `m.getArguments` (returns raw list)
  - **New**: `m.getInstanceLocation()` (returns nullable, needs `.toString()`), `m.getArguments` requires `.map(_.toString)`

  ### 3. Validation Error Format Changes

  ⚠️ **Breaking changes in error output format:**

  | Aspect | Old Format (1.0.76) | New Format (1.5.8) |
  |--------|---------------------|-------------------|
  | Path format | `$.field` | `/field` |
  | Array paths | `$.array[0]` | `/array/0` |
  | Path standard | JSONPath | JSON Pointer (RFC 6901) |
  | Error wording | "may only be 3 characters long" | "must be at most 3 characters long" |
  | Validation args | Limited context | Includes actual values |

  **Example:**
  ```scala
  // Old: "$.address: does not match the ipv4 pattern ^(...regex...)$"
  // New: "/address: does not match the ipv4 pattern must be a valid RFC 2673 IP address"
```

  ###  4. Test Updates

  All test assertions updated to match new error formats.

  ### 5. Build Tool Upgrade

  Upgraded sbt from 1.8.2 to 1.11.7 to support migration to Sonatype's Central Portal.

  **⚠️ Important Note:** Sonatype has sunset the Legacy OSSRH endpoint for publishing on 2025-06-30. To continue publishing to the Central Repository, sbt 1.11.x or later and sbt-ci-release 1.11.0 or later are required.

  ## Backward Compatibility

  ✅ Public API: CirceValidator's public API remains unchanged

  ⚠️ Error Messages: **Applications parsing error messages or paths will need updates**

  ## Release Notes Analysis

  Release notes from 1.0.76 to 1.5.8 have been analyzed. Key findings:
  - API modernization (builder patterns, better null safety)
  - Improved error messages with JSON Pointer standard
  - Enhanced Jackson compatibility
  - No identified issues that would harm the application
  - Security model maintained (no-op schema loader still prevents network calls)

### CODE REVIEW

File: modules/core/src/test/scala/com.snowplowanalytics.iglu.client/validator/CachingValidationSpec.scala [hunk 1]
```diff
@@ -109,8 +109,8 @@ class CachingValidationSpec extends Specification {
     val nonStringExpected = ValidatorError.InvalidData(
       NonEmptyList.of(
         ValidatorReport(
-          "$.country: integer found, string expected",
-          Some("$.country"),
+          "/country: integer found, string expected",
```
istreeter: This looks like a very significant change for Snowplow.

Am I right that this validator report is what gets returned to our customers via bad rows?  And therefore what gets displayed to users in the snowplow console?

I am a big fan of upgrading this library to the latest version.  But if we press ahead with this change, then we need to flag this change in behaviour to our Product team and possibly other Engineering teams too.

I assume there's no option to keep the old style validation message??

oguzhanunlu: replying for the record -- I pushed new commits to bring back legacy json path but there is nothing I can do about message wording, it is updated as static content

gthomson31:
## Build Tool Updates

I've added commits to upgrade the build tooling to support publishing to Sonatype's Central Portal:

### Changes Made

1. **sbt upgrade** (`1.8.2` → `1.11.7`)
   - Required for Central Portal migration
   - Sonatype has sunset the Legacy OSSRH endpoint as of 2025-06-30

2. **sbt-scoverage upgrade** (`2.0.4` → `2.4.0`)
   - Fixes compatibility with sbt 1.11.7
   - Version 2.4.0 includes support for Scala 2.12.20 and fallback support for older versions

3. **sbt-ci-release upgrade** (`1.5.7` → `1.9.2`)
   - Required version 1.11.0+ for Central Portal publishing
   - Updated artifact coordinates from `com.geirsson` to `com.github.sbt` (plugin changed maintainers in v1.6.0)

4. **Scala version upgrades** (for SIP-51 compliance with sbt 1.11.7)
   - Scala 2.13: `2.13.9` → `2.13.16`
   - Scala 2.12: `2.12.17` → `2.12.19`
   - kind-projector: `0.13.2` → `0.13.3`

### Why This Matters

Starting June 30, 2025, the Legacy OSSRH endpoint will no longer be available for publishing. These upgrades ensure that future releases can be published to the Central Repository through the new Central Portal.

sbt 1.11.7 enforces [SIP-51](https://docs.scala-lang.org/sips/drop-stdlib-forwards.html) (backwards-only binary compatibility), which requires the Scala compiler version to match or exceed the scala-library version on the classpath. The upgraded dependencies (particularly sbt-scoverage 2.4.0) pull in scala-library 2.13.16+, necessitating the Scala version upgrades.

### Testing

✅ All changes have been tested locally:
- `sbt update` - All dependencies resolve successfully
- `sbt compile` - All modules compile without errors across all Scala versions
- `sbt "+ test"` - All 121 tests pass across Scala 2.12.19/2.12.20, 2.13.16, and 3.2.0

**References:**
- [Sonatype Central Portal FAQ - Migration from Legacy OSSRH](https://central.sonatype.org/faq/what-is-different-between-central-portal-and-legacy-ossrh/#self-service-migration)
- [sbt-ci-release 1.11.0+ requirement](https://github.com/sbt/sbt-ci-release#migrating-from-legacy-sonatype-ossrh)
- [SIP-51: Drop Forwards Binary Compatibility of the Scala 2.13 Standard Library](https://docs.scala-lang.org/sips/drop-stdlib-forwards.html)

The CI build should now pass with these updates.

gthomson31:
## Update: Additional Changes for Central Portal Support

Following initial testing, I've made additional updates to complete the Sonatype Central Portal migration:

### Additional Plugin Upgrade
- **sbt-ci-release**: `1.9.2` → `1.11.2` (adds full Central Portal publishing support)
  - Version 1.11.0+ includes complete integration with the new Sonatype Central Portal API
  - Resolves authentication handling for new credential format
  - Reference: https://github.com/sbt/sbt-ci-release/releases/tag/v1.11.0

### Credentials Updated
- Updated GitHub Actions secrets (`SONA_USER`, `SONA_PASS`) with new "Maven Central - Deployment Key" credentials
- New credentials retrieved from Keeper and applied to repository secrets
- Required for authentication with Sonatype Central Portal (old Legacy OSSRH credentials no longer work)

### Testing Status
- ✅ Local builds: All tests pass (121 tests across all Scala versions)
- ✅ Formatting, binary compatibility, and publish checks: All pass
- 🔄 Release workflow: Re-triggered with tag 4.1.0-M1 to validate Central Portal publishing with new credentials

---

**Original comment summary:**
- sbt: `1.8.2` → `1.11.7`
- sbt-scoverage: `2.0.4` → `2.4.0`
- sbt-ci-release: `1.5.7` → `1.11.2` (updated from 1.9.2)
- Scala 2.13: `2.13.9` → `2.13.16`
- Scala 2.12: `2.12.17` → `2.12.19`
- kind-projector: `0.13.2` → `0.13.3`
- Workflow optimizations: Matrix strategies for parallel CI and releases

gthomson31:
## Credential Rollout Across Repositories

**Note**: IT Services will roll out the new Maven Central credentials to all repositories publishing to Maven Central.

📋 **Tracking ticket**: [ITS-1792](https://snplow.atlassian.net/browse/ITS-1792)

This ticket documents the process to:
- Identify all repositories using Maven Central publishing
- Update `SONA_USER` and `SONA_PASS` secrets with new "Maven Central - Deployment Key" credentials from Keeper
- Apply updates via Vault and Terraform
- Validate credentials work with Sonatype Central Portal

All Snowplow repositories publishing to Maven Central will need similar build tool upgrades (sbt 1.11.x, sbt-ci-release 1.11.0+, Scala version updates per SIP-51) to work with the new Central Portal endpoint.

[ITS-1792]: https://snplow.atlassian.net/browse/ITS-1792?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CustomMessageFormatter.scala
```diff
@@ -0,0 +1,108 @@
...
+        // This matches the validator library's ValidationMessage.Builder.build() logic
+        val processedArguments = transformArguments(validatorType, arguments)
+        val args               = (path +: processedArguments).map(_.asInstanceOf[AnyRef]).toArray
+        formatter.format(args)
+
+      case None =>
+        // Fallback for unknown validator types
+        // This provides a reasonable default while alerting us to any gaps
+        val argsStr = if (arguments.nonEmpty) s" (${arguments.mkString(", ")})" else ""
+        s"$path: validation failed for $validatorType$argsStr"
```
istreeter: Maybe it would be neater if you don't handle the `None` case here.  Instead let this function return a `Option[String]`.  And then you can call fall back to the original message:

```scala
CustomMessageFormatter.formatMessage(vt, p, a).getOrElse(m.getMessage)
```

What do you think?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CustomMessageFormatter.scala
```diff
@@ -0,0 +1,108 @@
...
+   * @return Transformed arguments matching 1.0.76 format
+   */
+  def transformArguments(validatorType: String, arguments: Seq[String]): Seq[String] = {
+    validatorType match {
+      // enum: newer versions quote the values like ["foo", "bar"] but 1.0.76 produced [foo, bar]
+      // See https://github.com/networknt/json-schema-validator/pull/1095
+      case "enum" if arguments.nonEmpty =>
+        arguments.map { arg =>
+          // Remove quotes from string values in JSON array: ["foo", "bar"] -> [foo, bar]
+          arg.replaceAll("\"([^\"]+)\"", "$1")
```
istreeter: You must be able to think of some input strings for which this replaceAll fails!

Clue: What if an enum string itself contains a `"`.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CustomMessageFormatter.scala
```diff
@@ -0,0 +1,108 @@
...
+   */
+  def formatMessage(validatorType: String, path: String, arguments: Seq[String]): String = {
+    messageTemplates.get(validatorType) match {
+      case Some(template) =>
+        val formatter = new MessageFormat(template)
+        // Build arguments array: [path, arg1, arg2, ...]
+        // This matches the validator library's ValidationMessage.Builder.build() logic
+        val processedArguments = transformArguments(validatorType, arguments)
+        val args               = (path +: processedArguments).map(_.asInstanceOf[AnyRef]).toArray
+        formatter.format(args)
```
istreeter: Is there any danger this could go wrong?  E.g. a future library upgrade, in which the args are no longer the expected length to match the template?

I know we have unit tests for most cases.... but is there a danger a template/args combination happens at run time that wasn't tested in the unit tests.

In other words, I'm wondering if we need a try/catch around this line.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CustomMessageFormatter.scala
```diff
@@ -0,0 +1,108 @@
...
+   * @return Formatted error message matching 1.0.76 format
+   */
+  def formatMessage(validatorType: String, path: String, arguments: Seq[String]): String = {
+    messageTemplates.get(validatorType) match {
+      case Some(template) =>
+        val formatter = new MessageFormat(template)
+        // Build arguments array: [path, arg1, arg2, ...]
+        // This matches the validator library's ValidationMessage.Builder.build() logic
+        val processedArguments = transformArguments(validatorType, arguments)
+        val args               = (path +: processedArguments).map(_.asInstanceOf[AnyRef]).toArray
```
istreeter: You can do this without mapping a list to a new list.

```scala
val args = (path +: processedArguments).asInstanceOf[List[AnyRef]].toArray

File: modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CustomMessageFormatter.scala [hunk 6]
```diff
@@ -0,0 +1,108 @@
...
+   * Message templates from json-schema-validator 1.0.76.
+   * Format uses Java MessageFormat placeholders: {0}, {1}, {2}, etc.
+   *
+   * Template arguments:
+   * - {0} is always the JSON path (e.g., "$.field" or "$")
+   * - {1}, {2}, ... are validator-specific arguments
+   *
+   * Loaded from jsv-messages-1.0.76.properties (43 templates total)
+   */
+  private val messageTemplates: Map[String, String] = {
```
istreeter: How about making it a `Map[String, MessageFormat]`?  i.e. compile the templates one time only.

And maybe make it a lazy val?  iglu-scala-client is sometimes used for fetching schemas, but not used for validation.  E.g. see many of our loaders.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
```diff
@@ -119,16 +121,54 @@ object CirceValidator extends Validator[Json] {
...
+    val rawArguments  = Option(m.getArguments).map(_.toList.map(_.toString)).getOrElse(List.empty)
+    val customMessage = getCustomMessage(m)
+
+    // Transform arguments to match 1.0.76 format for consistency with message formatting
+    val transformedArguments = Option(m.getType) match {
+      case Some(validatorType) =>
+        CustomMessageFormatter.transformArguments(validatorType, rawArguments).toList
+      case None =>
+        rawArguments
+    }
```
istreeter: This is all fine.  But...

It has a lot of edge case handling, like repeated checks for if `m.getType` is null, or if `m.getArguments` is null.  I feel like the code could be shuffled around a bit, to make it easier for a person to read to understand how all those edge cases are handled.  Something like:

```scala
Option(m.getType) match {
  case Some(validatorType) =>
    val customMessage = getCustomMessage(m, validatorType)
    val transformedArguments = CustomMessageFormatter.transformArguments(validatorType, rawArguments).toList
    ValidatorReport(
      customMessage,
      Option(m.getInstanceLocation()).map(_.toString),
      transformedArguments,
      Some(validatorType)
    )

  case None =>
    ValidatorReport(
      m.getMessage,
      Option(m.getInstanceLocation()).map(_.toString),
      rawArguments,
      None
    )
}
```

You might find other ways to make it neater for somebody reading the code.

modules/core/src/test/scala/com.snowplowanalytics.iglu.client/validator/ValidatorMessageFormatSpec.scala
```diff
@@ -0,0 +1,375 @@
...
+/**
+ * Comprehensive test spec to ensure all validator types maintain consistent
+ * error message formatting across json-schema-validator library versions.
+ *
+ * This spec tests validator types defined in CustomMessageFormatter to ensure:
+ * 1. Message templates match expected format from version 1.0.76
+ * 2. Argument positions ({0}, {1}, {2}) are correct
+ * 3. If the library changes behavior, these tests will fail
+ */
+class ValidatorMessageFormatSpec extends Specification {
```
istreeter, line 27: I have an idea.... is this possible?

We could merge this Spec into the develop branch as a separate PR.  Then we have proof that these tests pass against the develop branch.

Next, we rebase _this_ PR on top, and prove that this PR does not change the messages.

istreeter (commented):
Well done @oguzhanunlu.  I'm starting to believe this could work :)

modules/core/src/test/scala/com.snowplowanalytics.iglu.client/validator/ValidatorMessageFormatSpec.scala
oguzhanunlu: Great idea, let me create a PR with this spec only firstI

I added current edge case handling due to failing tests in CachingValidationSpec which showed that this PR (before edge case handling) changed messages for those cases. Your idea takes this confidence even further, thanks!

oguzhanunlu: I opened https://github.com/snowplow/iglu-scala-client/pull/266 where first commit confirms this spec matches 1.0.76 behavior. I also added a 2nd commit where I make sure we cover all json schema v4 types.

oguzhanunlu:
After merging #266 , I rebased on top of develop (required force-push), made sure 1.5.8 upgrade doesn't change error message for any type. I addressed feedback in a separate commit, along with an optimization `alreadyTransformed`.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CustomMessageFormatter.scala
oguzhanunlu: I agree and addressed in a new commit

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CustomMessageFormatter.scala
```diff
@@ -0,0 +1,180 @@
...
+          // Transform "$.not" : {...} to "not" : {...}
+          arg.replaceFirst("\"\\$\\.not\"", "\"not\"")
+        }
+
+      // oneOf: 1.5.8 passes ["2", "0, 1"] (count and matching indices)
+      // but 1.0.76 passed ["{\"type\":\"string\"}{\"type\":\"number\"}"] (concatenated schemas)
+      // We extract schemas at indices 0 and 1 from schemaNode and concatenate their JSON
+      case "oneOf" if arguments.size >= 2 && schemaNode.exists(_.isArray) =>
+        val indicesStr = arguments(1)
+        val indices    = indicesStr.split(",\\s*").map(_.trim.toInt).toList
```
istreeter: `.toInt` can throw if the string is something unexpected.  Do you think we should try/catch here?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
```diff
@@ -119,16 +129,67 @@ object CirceValidator extends Validator[Json] {
...
+    // Transform arguments once to match 1.0.76 format, reuse for both message and storage
+    val transformedArguments = Option(m.getType) match {
+      case Some(validatorType) =>
+        CustomMessageFormatter.transformArguments(validatorType, rawArguments, schemaNode).toList
+      case None =>
+        rawArguments
+    }
+
+    // Format message using already-transformed arguments
+    val customMessage = (Option(m.getType), transformedArguments) match {
```
istreeter: I don't see why you include `transformedArguments` in the pattern match here.  Maybe I'm missing something?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
```diff
@@ -119,16 +129,67 @@ object CirceValidator extends Validator[Json] {
       }
     } yield ()
 
-  private def fromValidationMessage(m: ValidationMessage): ValidatorReport =
-    ValidatorReport(m.getMessage, m.getPath.some, m.getArguments.toList, m.getType.some)
+  /**
+   * Extract custom formatted message from ValidationMessage.
+   * This is used for both data validation errors and schema validation errors.
```
istreeter: Is this line true?  I think this function is only used for validating schemas?

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
```diff
@@ -119,16 +129,67 @@ object CirceValidator extends Validator[Json] {
...
+   */
+  private def getCustomMessage(m: ValidationMessage): String = {
```
istreeter: I wonder if this whole function could be replaced with:

```scala
private def getCustomMessage(m: ValidationMessage): String =
  fromValidationMessage(m).message
```

I know that's a bit wasteful in a way, because it calculates all fields of the `ValidatorReport` and then ignores most of those fields.  But, I _think_ this method is only called when validating schemas.  And only if a schema is invalid.

When you think about it... it's quite rare this function will ever get executed.  We don't expect our pipelines to encounter many invalid schemas ever!  So I think simplicity is more important than an an optimized implementation for this function.

Even better, if you pick this approach then you could remove `alreadyTransformed` from `CustomMessageFormatter.formatMessage`.  That removes even more complexity.

Please check my thinking before agreeing!

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CustomMessageFormatter.scala
```diff
@@ -0,0 +1,180 @@
...
+    schemaNode: Option[JsonNode] = None
+  ): Seq[String] = {
+    validatorType match {
+      // enum: 1.5.8 quotes the values like ["foo", "bar"] but 1.0.76 produced [foo, bar]
+      // See https://github.com/networknt/json-schema-validator/pull/1095
+      case "enum" if arguments.nonEmpty =>
+        arguments.map { arg =>
+          // Use Jackson to parse the JSON array properly and rebuild without quotes
+          Try {
+            val arrayNode = objectMapper.readTree(arg)
```
istreeter, line 93: This is just an idea.  I haven't thought it through fully........

On [this line](https://github.com/snowplow/iglu-scala-client/blob/9208076767261cd1c7e6b7f1762c27e0f0f5b98b/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala#L156) you convert all the arguments into strings.  And therefore on this line here you need to parse those strings again.

I wonder if there is an alternative implementation in which you don't convert the arguments to strings.  If you pass the arguments around as `Seq[AnyRef]` then you won't need the `objectMapper` in this function.

Like I say, that's only an idea, I'm not saying definitely do it.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CustomMessageFormatter.scala
oguzhanunlu: It is supposed to be an index but it could be an invalid int in future in case library changes behavior which would surface in test failures but I'll add a try-catch here to be defensive. In case at least one index is invalid in future, oneOf test case will fail and this will tell us about the library change as well.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
oguzhanunlu: comment was outdated, I agree with your point and I'll prefer simplicity 👍

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CustomMessageFormatter.scala
oguzhanunlu: Good idea but lib doesn't give us JsonNode, it gives us a pre-formatted string e.g. `["red", "green", "blue"]` as a single argument, so I don't really convert from JsonNode to String at the beginning. We have to parse this string to rebuild without quotes, so I need objectMapper.

istreeter:
I think this PR is starting to look very neat.  I'm pleased you've found a way we can move to the latest version of the 3rd party library without causing migration pain for our platform

BUT.... I think we need to ask the opinion of others in our team.  This PR has gone way beyond the scope of what we planned this issue got discussed in the team meeting.  The team will be responsible for maintaining this newly added code.  I don't want to approve this PR unless everyone is on board with taking on this new code.

spenes (approved):
It adds fair bit of complexity but I guess not much to do preserve backward compatibility with downstream tools.

PR #269, opened 2025-12-04
Create new public method CirceValidator#compileJsonSchema

gleb-lobov:
Would clean up [this](https://github.com/snowplow/enrich-private/pull/151) enrich PR

### DIFF
--- a/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
+++ b/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
@@ -115,8 +115,8 @@ object CirceValidator extends Validator[Json] {
     }
   }
 
-  /** Validate instance against schema and return same instance */
-  private def validateOnReadySchema(
+  /** Validate instance against schema */
+  def validateOnReadySchema(
     schema: JsonSchema,
     instance: Json,
     maxJsonDepth: Int
@@ -165,7 +165,8 @@ object CirceValidator extends Validator[Json] {
     )
   }
 
-  private def evaluateSchema(
+  /** Compile jackson schema to networknt schema */
+  def evaluateSchema(
     schemaAsNode: JsonNode
   ): Either[ValidatorError.InvalidSchema, JsonSchema] = {
     Either
--- a/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
+++ b/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
@@ -165,8 +165,7 @@ object CirceValidator extends Validator[Json] {
     )
   }
 
-  /** Compile jackson schema to networknt schema */
-  def evaluateSchema(
+  private def evaluateSchema(
     schemaAsNode: JsonNode
   ): Either[ValidatorError.InvalidSchema, JsonSchema] = {
     Either
@@ -189,6 +188,26 @@ object CirceValidator extends Validator[Json] {
             fromValidationMessage(m).message
           )
       )
+  }
+
+  /** Compile circe json to networknt schema */
+  def compileJsonSchema(
+    schema: Json,
+    maxJsonDepth: Int
+  ): Either[ValidatorError.InvalidSchema, JsonSchema] = {
+    for {
+      schemaAsNode <- circeToJackson(schema, maxJsonDepth).leftMap(_.toInvalidSchema)
+      _            <- validateSchema(schemaAsNode)
+      evaluated    <- evaluateSchema(schemaAsNode)
+    } yield evaluated
+  }
+
+  private def validateSchema(schema: JsonNode): Either[ValidatorError.InvalidSchema, Unit] = {
+    val issues = validateSchemaAgainstV4(schema)
+    issues match {
+      case Nil          => Right(())
+      case head :: tail => Left(ValidatorError.InvalidSchema(NonEmptyList(head, tail)))
+    }
   }
 
   private[client] object WithCaching {
@@ -231,42 +250,14 @@ object CirceValidator extends Validator[Json] {
             case Some(alreadyEvaluatedSchema) =>
               alreadyEvaluatedSchema.pure[F]
             case None =>
-              provideNewJsonSchema(schema, maxJsonDepth)
+              compileJsonSchema(schema, maxJsonDepth)
                 .pure[F]
                 .flatTap(result => evaluationCache.put((key, timestamp), result))
           }
         case ResolverResult.NotCached(SchemaItem(schema, _)) =>
-          provideNewJsonSchema(schema, maxJsonDepth).pure[F]
+          compileJsonSchema(schema, maxJsonDepth).pure[F]
       }
     }
 
-    private def provideNewJsonSchema(
-      schema: Json,
-      maxJsonDepth: Int
-    ): Either[ValidatorError.InvalidSchema, JsonSchema] = {
-      for {
-        schemaAsNode <- circeToJackson(schema, maxJsonDepth).leftMap(_.toInvalidSchema)
-        _            <- validateSchema(schemaAsNode)
-        evaluated    <- evaluateSchema(schemaAsNode)
-      } yield evaluated
-    }
-
-    private def validateSchema(schema: JsonNode): Either[ValidatorError.InvalidSchema, Unit] = {
-      val issues = V4Schema
-        .validate(schema)
-        .asScala
-        .toList
-        .map(m =>
-          ValidatorError.SchemaIssue(
-            Option(m.getInstanceLocation()).map(_.toString).getOrElse(""),
-            fromValidationMessage(m).message
-          )
-        )
-
-      issues match {
-        case Nil          => Right(())
-        case head :: tail => Left(ValidatorError.InvalidSchema(NonEmptyList(head, tail)))
-      }
-    }
   }
 }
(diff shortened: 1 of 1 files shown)


istreeter:
@gleb-lobov I don't quite understand why **these** are the two methods you want to make public.

Comparing with the Enrich PR snowplow/enrich-private#151, it seems to me the functionality you need to import/replace is [this part](https://github.com/snowplow/enrich-private/blob/event-spec-enrichment-impl/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EventSpecEnrichment.scala#L184-L193) where you convert from circe Json into networknet JsonSchema.  So the corresponding function in iglu-scala-client is `provideNewJsonSchema`, [this function](https://github.com/snowplow/iglu-scala-client/blob/4.0.3/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala#L242-L251).

gleb-lobov:
@istreeter I was following the public method `CirceValidator#validate`, and it never calls `provideNewJsonSchema`. `provideNewJsonSchema` is part of the private object, and is called from the `CirceValidator.WithCaching#validate`, that is used internally to cache things. Since we are doing our own caching, and exposing a private object is awkward, I've just made public the method used in the main object.

I can move the `provideNewJsonSchema` to the main object and make is public, if you see it fit. Seems that it does not rely on caching and logically can belong to the main object

istreeter:
> I was following the public method `CirceValidator#validate`

OK I think I get what you're saying.

In your Enrich PR snowplow/enrich-private#151 there are two different places where you have written code which could be replaced by iglu-scala-client:

**1. Validating an entity against the schema**

[This function here](https://github.com/snowplow/enrich-private/blob/a237fc85a0481b685c4bc41fd0eb067b4746784e/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EventSpecEnrichment.scala#L195-L202) where you validate the entity against the schema.  I agree that `CirceValidator#validateOnReadySchema` gives you the functionality you need to replace the implementation in Enrich.

But... I don't think that implementation is a problem in Enrich.  I mean, it's just two lines of code.  I don't think you benefit much from importing the implementation from iglu-scala-client

```scala
    (for {
      jacksonJson <- circeToJackson(data, maxJsonDepth).toOption
    } yield schema.validate(jacksonJson).isEmpty).getOrElse(false)
```

**2. Compiling the json schema**

[This function here](https://github.com/snowplow/enrich-private/blob/a237fc85a0481b685c4bc41fd0eb067b4746784e/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EventSpecEnrichment.scala#L184-L193) where you compile the json schema.  In this case the Enrich implementation adds a lot of overhead.  e.g. you were forced to re-define `SchemaValidatorsConfig` and `IgluMetaschemaFactory` and `fakeUrlFetcher` etc.

So **this** is the functionality which I think we should look to import from iglu-scala-client.  And this is why I think you need `provideNewJsonSchema` to be public.

gleb-lobov:
>1. Validating an entity against the schema

Yeah, I can skip it, but it would be nice to not even call `circeToJackson`. But I can skip making `validateOnReadySchema` public if you don't think it makes sense.

>2. Compiling the json schema

You are absolutely right that this is more problematic, and I have to bring a lot. My idea was that if the `evaluateSchema` is public and `validateOnReadySchema` is public, the implementation for my [compileJsonSchema](https://github.com/snowplow/enrich-private/blob/a237fc85a0481b685c4bc41fd0eb067b4746784e/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EventSpecEnrichment.scala#L184C15-L184C32) would be just
```scala
  def compileJsonSchema(data: Json, schema: Json): Either[ValidatorError, Unit] =
    for {
      jacksonJson <- circeToJackson(schema, Int.MaxValue).leftMap(_.toInvalidSchema)
      schema      <- evaluateSchema(jacksonJson)
      _           <- validateOnReadySchema(schema, data, Int.MaxValue)
    } yield ()

```

that is just a copy of a [CirceValidator#validate](https://github.com/snowplow/iglu-scala-client/blob/master/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala#L100-L105)

I think the confusion is because Im looking at [CirceValidator#validate](https://github.com/snowplow/iglu-scala-client/blob/master/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala#L100-L105) and you are looking at [CirceValidator.WithCaching#validate](https://github.com/snowplow/iglu-scala-client/blob/master/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala#L206-L219)

Form the looks of it, I don't see any difference, since we'll not use caching. Let me know if I should make `provideNewJsonSchema` public instead

istreeter:
But for _compiling_ the jsonschema you cannot have this signature

```scala
def compileJsonSchema(data: Json, schema: Json): Either[ValidatorError, Unit]
```

...because you don't have the `data` yet.  The `data` comes later when you start enriching events.  Whereas you compile the jsonschema during application start up.

gleb-lobov:
Yeah, you're right, sorry, blind copy-paste

This is what I've meant(ignore the left channel, just converting everything to string so it compiles)
```scala
  def compileJsonSchema(circe: Json): Either[String, JsonSchema] =
    for {
      jacksonJson <- circeToJackson(circe, Int.MaxValue).leftMap(_.message)
      schema      <- evaluateSchema(jacksonJson).leftMap(_.toString)
    } yield schema
```

istreeter:
~But I don't think that works either, because in your example `schema` is a jackson `JsonNode`, whereas you need to compile to a networknt `JsonSchema`.~

Sorry ignore that comment I wrote the wrong thing.

istreeter:
OK I understand what you're saying now.  So `evaluateSchema` and `provideNewSchema` are very similar functions.  They both have the same return type.  The former accepts a jackson json, whereas the latter accepts a circe json.  From Enrich point of view I think we could make either one public.  But from point of view of designing the public API of iglu-scala-client, I _think_ it makes slightly more sense to expose a function that accepts a circe json.

And from point of view of designing the public API... it should probably live in the top-level object (like you said) and have a name like `compileJsonSchema`.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/validator/CirceValidator.scala
```diff
@@ -190,6 +190,26 @@ object CirceValidator extends Validator[Json] {
...
+  ): Either[ValidatorError.InvalidSchema, JsonSchema] = {
+    for {
+      schemaAsNode <- circeToJackson(schema, maxJsonDepth).leftMap(_.toInvalidSchema)
+      _            <- validateSchema(schemaAsNode)
+      evaluated    <- evaluateSchema(schemaAsNode)
+    } yield evaluated
+  }
+
+  private def validateSchema(schema: JsonNode): Either[ValidatorError.InvalidSchema, Unit] = {
+    val issues = validateSchemaAgainstV4(schema)
```
gleb-lobov, line 205: This was exact same as `validateSchemaAgainstV4` in the previous implementation, so I've just used the function.

gleb-lobov:
@istreeter yeah, I see now what you mean. I've moved the `provideNewSchema` to main object, made it public, and renamed to `compileJsonSchema`. I've also moved the `validateSchema` to the main object, since it has nothing to do with caching. I've also changed the implementation of it to re-use the `validateSchemaAgainstV4`. 

The `WithCaching` now looks clean to me, as it only deals with the caching logic. Let me know if the public API also looks fine from your side

istreeter:
Thanks @gleb-lobov I think the `compileJsonSchema` looks perfect now.

Regarding `validateOnReadySchema` I still lean towards **not** making this one public.  My reason is, `validateOnReadySchema` validates the data AND then it calls `fromValidationMessage` to convert from `ValidationMessage` to `ValidatorReport`.  This seems wasteful because:

1. In your Enrich PR you don't need the `ValidatorReport`, you only need to know if the result is valid or invalid.  i.e. you just call `.isEmpty` [on this line](https://github.com/snowplow/enrich-private/blob/a237fc85a0481b685c4bc41fd0eb067b4746784e/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EventSpecEnrichment.scala#L202).
2. `fromValidationMessage` does some string manipulation, so it is not completely cheap
3. For event spec inference I guess we expect lots of data to be invalid.  So we will hit this `fromValidationMessage` quite often.  Because the whole point of _inference_ is we don't know in advance if some data is expected to validate.

I think your function `isValidAgainstSchema` [in your Enrich PR](https://github.com/snowplow/enrich-private/blob/event-spec-enrichment-impl/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EventSpecEnrichment.scala#L195-L202) is fine already.  It is concise and easy to read.  I don't feel like you need to shift the implementation over to iglu-scala-client.

gleb-lobov:
@istreeter good note about the error string manipulation, I didn't think about this. 

I've made the `validateOnReadySchema` private, you are correct that the `compileJsonSchema` is enough to make enrich PR look decent, and we also save up some performance by doing the `isValidAgainstSchema` the way it is now

istreeter (approved):
This all looks good now.  I will merge this into develop for you, and I will tag a pre-release version which you can import into Enrich.  We can turn it into a proper release after everything is tested in Enrich.

istreeter:
@gleb-lobov 4.1.0-M2 is now published to Sonatype, so you can import it into Enrich.

PR #270, opened 2025-12-17
Add retry mechanism to idLookupInstance

spenes:
Jira ref: [PDP-2269](https://snplow.atlassian.net/browse/PDP-2269)

[PDP-2269]: https://snplow.atlassian.net/browse/PDP-2269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/JavaNetRegistryLookup.scala
+++ b/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/JavaNetRegistryLookup.scala
@@ -32,7 +32,10 @@ import scala.util.control.NonFatal
 
 object JavaNetRegistryLookup {
 
-  private val ReadTimeoutMs = 4000L
+  private val ReadTimeoutMs    = 4000L
+  private val MaxRetries       = 5
+  private val InitialBackoffMs = 100L
+  private val MaxBackoffMs     = 5000L
 
   private lazy val httpClient = HttpClient
     .newBuilder()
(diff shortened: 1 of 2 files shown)


istreeter (commented):
OK this PR does more than I expected.  Which is both good and (possibly) bad.  Maybe not bad, but needs thinking.

This Jira ticket said to change the `idLookupInstance`.  But here you change the `idLookupInstance` AND the `ioLookupInstance`.  In changing the io instance, you also changed `Sync` to `Async` in the function arguments.

That _might_ be ok.  But we normally try to maintain api compatibility between minor versions.  Very strictly speaking, I think it is technically a breaking change to change `Sync` to `Async`.  In reality it is probably fine because the IO `Sync` instance is also a `Async`.  And furthermore we don't use `ioLookupInstance` in production so we are unlikely to break any snowplow app.

What do you think?  We can either....

- Merge this.  Knowing it is _probably_ safe to change `Sync` to `Async`.  And if we're wrong then probably a QA test will fail somewhere.
- Or, revert the change to `ioLookupInstance`, and keep the change to `idLookupInstance`.

Am I being overly paranoid?

Apart from this doubt, I like the PR and I think you implemented it well.

spenes:
Thanks @istreeter, let's be cautious and revert the change to `ioLookupInstance`. I will revert it.

istreeter (approved):

PR #273, opened 2025-12-26
Add Resolver#isSystemError

oguzhanunlu:
Add Resolver#isSystemError to allow users to determine if a resolution error is caused by Iglu server unavailability.

Returns true when:
- At least one custom Iglu Server is *unhealthy*
- All Iglu Central mirrors are *unhealthy*

`*unhealthy*`: errors of a LookupHistory can be either
- full of RepoFailure: easy to consider this registry as unhealthy
- a mix of RepoFailure along with NotFound and/or ClientFailure: I considered the registry unavailable, since errors is a Set[RegistryError] which is unordered
- no RepoFailure at all: can't consider registry as unhealthy

Returns false when:
- At least one Iglu Central mirror is healthy
- All custom Iglu Servers are healthy

ref: https://snplow.atlassian.net/browse/PDP-2318

### DIFF
--- a/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
+++ b/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
@@ -206,28 +206,30 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
   }
 
   /**
-   * Checks if the resolution error indicates server unavailability.
+   * Checks if the resolution error indicates a fatal/unrecoverable issue.
    *
    * Returns true when:
-   * - Any custom (non-Iglu Central) repository has a RepoFailure
-   * - All Iglu Central mirrors have a RepoFailure
+   * - Any custom (non-Iglu Central) repository has a RepoFailure or ClientFailure
+   * - All Iglu Central mirrors have a RepoFailure or ClientFailure
    *
    * Returns false otherwise.
    *
-   * If a registry has RepoFailure alongside other errors, assume it is unavailable.
+   * If a registry has a fatal error alongside other errors, assume it is unrecoverable.
    */
   def isUnrecoverable(error: ResolutionError): Boolean = {
     val (igluCentral, custom) = error.value.partition { case (repo, _) =>
       allIgluCentral.contains(repo)
     }
 
-    def hasRepoFailure(history: LookupHistory): Boolean =
-      history.errors.exists(_.isInstanceOf[RegistryError.RepoFailure])
+    def hasFatalError(history: LookupHistory): Boolean =
+      history.errors.exists(e =>
+        e.isInstanceOf[RegistryError.RepoFailure] || e.isInstanceOf[RegistryError.ClientFailure]
+      )
 
-    val customUnavailable = custom.values.exists(hasRepoFailure)
+    val customUnavailable = custom.values.exists(hasFatalError)
 
     val igluCentralUnavailable =
-      igluCentral.nonEmpty && igluCentral.values.forall(hasRepoFailure)
+      igluCentral.nonEmpty && igluCentral.values.forall(hasFatalError)
 
     customUnavailable || igluCentralUnavailable
   }
--- a/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
+++ b/modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
@@ -206,6 +206,33 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
   }
 
   /**
+   * Checks if the resolution error indicates server unavailability.
+   *
+   * Returns true when:
+   * - Any custom (non-Iglu Central) repository has a RepoFailure
+   * - All Iglu Central mirrors have a RepoFailure
+   *
+   * Returns false otherwise.
+   *
+   * If a registry has RepoFailure alongside other errors, assume it is unavailable.
+   */
+  def isUnrecoverable(error: ResolutionError): Boolean = {
+    val (igluCentral, custom) = error.value.partition { case (repo, _) =>
+      allIgluCentral.contains(repo)
+    }
+
+    def hasRepoFailure(history: LookupHistory): Boolean =
+      history.errors.exists(_.isInstanceOf[RegistryError.RepoFailure])
+
+    val customUnavailable = custom.values.exists(hasRepoFailure)
+
+    val igluCentralUnavailable =
+      igluCentral.nonEmpty && igluCentral.values.forall(hasRepoFailure)
+
+    customUnavailable || igluCentralUnavailable
+  }
+
+  /**
    * The variant of lookupSchemasUntilResult that returns the result
    * that isn't wrapped with ResolverResult
    */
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -205,6 +205,33 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+   *
+   * If a registry has RepoFailure alongside other errors, assume it is unavailable.
+   */
+  def isUnrecoverable(error: ResolutionError): Boolean = {
+    val (igluCentral, custom) = error.value.partition { case (repo, _) =>
+      allIgluCentral.contains(repo)
+    }
+
+    def hasRepoFailure(history: LookupHistory): Boolean =
+      history.errors.exists(_.isInstanceOf[RegistryError.RepoFailure])
```
istreeter: There are three types of `RegistryError`.  Out of the three types, only `NotFound` is expected, i.e. a tracking error not a pipeline error.  Both of the other two (`RepoFailure` and `ClientError`) indicate some kind of issue with the pipeline, which should never happen.  How about we crash on `ClientError` too?  I cannot think of a reason not to.

Even the scaladoc for `ClientError` says [it is fatal](https://github.com/snowplow/iglu-scala-client/blob/4.1.0/modules/data/src/main/scala/com.snowplowanalytics.iglu/client/resolver/registries/RegistryError.scala#L31).

oguzhanunlu: yeah that makes sense, I'll push a new commit for that

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -205,6 +205,35 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+   *
+   * Returns true when:
+   * - Any custom (non-Iglu Central) repository has a RepoFailure or ClientFailure
+   * - All Iglu Central mirrors have a RepoFailure or ClientFailure
+   *
+   * Returns false otherwise.
+   *
+   * If a registry has a fatal error alongside other errors, assume it is unrecoverable.
+   */
+  def isUnrecoverable(error: ResolutionError): Boolean = {
```
istreeter: I have doubts about the name `isUnrecoverable`.  Bearing in mind this becomes part of the public API of this library, so name is quite important.

If `isUnrecoverable` returns `false` then it implies this error is "recoverable".  But what does "recoverable" mean in this context?  It means we're going to send this event to the bad stream... but I wouldn't say that is an example of "being recoverable".

You know I'm not good at naming things :) so I don't have good suggestions to change the name.

Maybe `isSystemError` or `isPlatformError`... meaning the system/platform/deployment/configuration is the problem, and there is no problem with the data.

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -205,6 +205,35 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+    val (igluCentral, custom) = error.value.partition { case (repo, _) =>
+      allIgluCentral.contains(repo)
+    }
+
+    def hasFatalError(history: LookupHistory): Boolean =
+      history.errors.exists(e =>
+        e.isInstanceOf[RegistryError.RepoFailure] || e.isInstanceOf[RegistryError.ClientFailure]
+      )
```
istreeter: This is fine.  My personal preference is to do something like:

```scala
history.errors.exists {
  case _: RegistryError.RepoFailure => true
  case _: RegistryError.ClientFailure => true
  case _: RegistryError.NotFound => false
}
```

...because that is more explicit to someone reading the code about how we categorize each of the error types.

But I don't want to nitpick, and other engineers will probably disagree with my preferences!  So how you did it is fine.

oguzhanunlu: > because that is more explicit to someone reading the code about how we categorize each of the error types

exactly, I like it and will modify it. Thanks for the suggestion 👍

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
oguzhanunlu: I think `isSystemError` sounds good, thanks for the suggestion!

modules/core/src/main/scala/com.snowplowanalytics.iglu/client/resolver/Resolver.scala
```diff
@@ -205,6 +205,37 @@ final case class Resolver[F[_]](repos: List[Registry], cache: Option[ResolverCac
...
+  /**
+   * Checks if the resolution error indicates a system/platform issue rather than a data issue.
+   *
+   * Returns true when:
+   * - Any custom (non-Iglu Central) repository has a RepoFailure or ClientFailure
+   * - All Iglu Central mirrors have a RepoFailure or ClientFailure
+   *
+   * Returns false otherwise (e.g., NotFound errors indicate missing schema, not system failure).
+   *
+   * If a registry has a fatal error alongside other errors, assume it is a system error.
```
istreeter: I just noticed this comment.  Please can you help me understand exactly what happens if a registry has fatal errors AND non-fatal errors.

Imagine a pipeline has some mis-configured tracking, so it generates a lot of `NotFound` schema violations.  In this scenario the system is healthy, even though the tracking is wrong.  This is probably a very common scenario.

Now imagine the Iglu Server goes down for this pipeline and starts returning 5xx for all requests.  I think this an example where the `LookupHistory` will contain a mixture of fatal errors (`RepoFailure`) AND what you call "other" errors (`NotFound`).

When you think about this scenario... we don't want Enrich to crash in this scenario.  We have successfully cached the fact that this schema is not found in all repos.  So I think it's better to generate a schema violation with `NotFound` instead of crashing Enrich.

So.... can you describe what happens in that scenario I described?  Would Enrich crash or stay alive in that scenario?  And if you need to update the implementation and tests, then please do so.

P.S. Sorry I keep noticing new things to complain about!!

oguzhanunlu: I'm glad that you question the approach early rather than late!

In your scenario, enrich would crash. Why? LookupHistory#errors is of type `Set[RegistryError]` which is unordered:

{NotFound, RepoFailure} could mean:
  - NotFound first → then server went down (RepoFailure) = server is currently down
  - RepoFailure first → then server recovered and said NotFound = server is healthy now

Since we can't distinguish, I went down the current route so that events wouldn't go to bad stream.

> So I think it's better to generate a schema violation with NotFound instead of crashing Enrich.

Yeah I agree. I put too much priority on the idea of `events shouldn't go to bad`. I'll change accordingly. Thanks for the heads up!

istreeter (approved):
I think this is fine.

We should get the opinion of @benjben before releasing Enrich, because he previously thought about Iglu errors when working on a similar feature.  He might spot something I missed.

It's fine to merge this for now.  If @benjben has any suggestions for improvement then we can address it in a follow-up PR.

benjben:
I think that the algorithm looks great !

If for a repo we have first a failure and then a `NotFound` it works well as `NotFound` takes precedence over any other kind of failure.

